PR #6, opened 2023-03-14
Update oasis3 mct package.py

harshula:


packages/oasis3-mct/package.py
```diff
@@ -180,18 +197,140 @@ def __create_pkgconfig(self, spec, prefix):
...
+        # TODO: Clean up the debug / non-debug mechanism once the compiler
+        #       flag audit is finished.
+        #       https://github.com/ACCESS-NRI/ACCESS-OM/issues/12
+        NCI_INTEL_FLAGS = "-r8 -i4 -traceback -fpe0 -convert big_endian -fno-alias -ip -check noarg_temp_created"
+        NCI_OPTIM_FLAGS = "-g3 -O2 -axCORE-AVX2 -debug all -check none -qopt-report=5 -qopt-report-annotate"
+        CFLAGS = ""
+        if "+nodebuginfo" in self.spec:
+            NCI_INTEL_FLAGS = "-r8 -i4 -fpe0 -convert big_endian -fno-alias -ip -check noarg_temp_created"
+            NCI_OPTIM_FLAGS = "-g0 -O0 -axCORE-AVX2 -debug none -check none -qopt-report=5 -qopt-report-annotate"
+            CFLAGS = "-g0"
```
aidanheerdegen, line 226:
```suggestion
        if "+nodebuginfo" in self.spec:
            NCI_OPTIM_FLAGS = " ".join(NCI_OPTIM_FLAGS, "-g0 -debug none")
            CFLAGS = "-g0"
        else:
            NCI_INTEL_FLAGS = " ".join(NCI_INTEL_FLAGS, "-traceback")            
            NCI_OPTIM_FLAGS = " ".join(NCI_OPTIM_FLAGS, "-g3 -debug all")
```
I'm suggesting setting the common flags in one location and just adding the debug option here. Maybe this is what your TODO above is referencing? 

Could you check I've got the common flags correct? I assumed the `-O0` at line 225 was a mistake?

packages/oasis3-mct/package.py
```diff
@@ -180,18 +197,140 @@ def __create_pkgconfig(self, spec, prefix):
...
+    # https://spack-tutorial.readthedocs.io/en/latest/tutorial_buildsystems.html
+    def edit(self, spec, prefix):
+        SRCDIR = self.stage.source_path
+        makeinc_path = join_path(SRCDIR, self.__makeinc)
+        config = {}
+
+        # TODO: Clean up the debug / non-debug mechanism once the compiler
+        #       flag audit is finished.
+        #       https://github.com/ACCESS-NRI/ACCESS-OM/issues/12
+        NCI_INTEL_FLAGS = "-r8 -i4 -traceback -fpe0 -convert big_endian -fno-alias -ip -check noarg_temp_created"
```
aidanheerdegen, line 220:
```suggestion
        NCI_INTEL_FLAGS = "-r8 -i4 -fpe0 -convert big_endian -fno-alias -ip -check noarg_temp_created"
```
I suggest reducing this to the options common to both variants and then adding the variant-specific flags later.

packages/oasis3-mct/package.py
```diff
@@ -180,18 +197,140 @@ def __create_pkgconfig(self, spec, prefix):
...
+        NCI_OPTIM_FLAGS = "-g3 -O2 -axCORE-AVX2 -debug all -check none -qopt-report=5 -qopt-report-annotate"
```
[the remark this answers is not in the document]
harshula: Please comment in https://github.com/ACCESS-NRI/ACCESS-OM/issues/12

packages/oasis3-mct/package.py
harshula: Please comment in https://github.com/ACCESS-NRI/ACCESS-OM/issues/12

packages/oasis3-mct/package.py
harshula: Please comment in https://github.com/ACCESS-NRI/ACCESS-OM/issues/12

packages/oasis3-mct/package.py
aidanheerdegen: Have done so but this

> Could you check I've got the common flags correct? I assumed the -O0 at line 225 was a mistake?

Is just a mistake I think. From https://github.com/ACCESS-NRI/ACCESS-OM/issues/12#issuecomment-1428923689 the `make.nci` has this:
```
ifeq ($(DEBUG), yes)
    NCI_DEBUG_FLAGS = -g3 -O0 -fpe0 -no-vec -debug all -check all -no-vec
    F90FLAGS_1      = $(NCI_INTEL_FLAGS) $(NCI_REPRO_FLAGS) $(NCI_DEBUG_FLAGS)
    CPPDEF          = -Duse_netCDF -Duse_comm_$(CHAN) -DTREAT_OVERLAY -DDEBUG -D__VERBOSE
    MCT_FCFLAGS     = $(NCI_REPRO_FLAGS) $(NCI_DEBUG_FLAGS) -ip
else
    NCI_OPTIM_FLAGS = -g3 -O2 -axCORE-AVX2 -debug all -check none -qopt-report=5 -qopt-report-annotate
    F90FLAGS_1      = $(NCI_INTEL_FLAGS) $(NCI_REPRO_FLAGS) $(NCI_OPTIM_FLAGS)
    CPPDEF          = -Duse_netCDF -Duse_comm_$(CHAN) -DTREAT_OVERLAY
    MCT_FCFLAGS     = $(NCI_REPRO_FLAGS) $(NCI_OPTIM_FLAGS) -ip
endif
```

Specifically:
```make
NCI_OPTIM_FLAGS = -g3 -O2 -axCORE-AVX2 -debug all -check none -qopt-report=5 -qopt-report-annotate
```
versus
```python
NCI_OPTIM_FLAGS = "-g0 -O0 -axCORE-AVX2 -debug none -check none -qopt-report=5 -qopt-report-annotate"
```

I understand some of those flags are changed for build repro reasons, but surely it should be `-O2` in this case?

packages/oasis3-mct/package.py
aidanheerdegen: I've made a comment in the above issue with respect to the flags being used, but this and the other comments above that concern layout don't depend on the flags used, they're about specifying information once: using a common base of options and adding to it rather than repeating the same information. 

I think it is better because it clearly delineates which options are **always** used, and which are dependent on optimisation/debugging/profiling etc.

packages/oasis3-mct/package.py
harshula: # TODO: Clean up the debug / non-debug mechanism once the compiler
        #       flag audit is finished.
        #       https://github.com/ACCESS-NRI/ACCESS-OM/issues/12

packages/oasis3-mct/package.py
```diff
@@ -7,33 +7,39 @@
...
+# https://spack.readthedocs.io/en/latest/build_systems/makefilepackage.html
+class Oasis3Mct(MakefilePackage):
     """ACCESS-NRI's fork of https://gitlab.com/cerfacs/oasis3-mct OASIS3-MCT 2.0."""
 
     homepage = "https://www.access-nri.org.au"
     git = "https://github.com/ACCESS-NRI/oasis3-mct.git"
     # NOTE: URL definition required for CI
     # Spack needs tarball URL to be defined to access github branches
-    url = "https://github.com/ACCESS-NRI/oasis3-mct/tarball/5e57b1aa40840adb6d6747d475dabadab5d9d6fb"
+    url = "https://github.com/ACCESS-NRI/oasis3-mct/tarball/spack-build"
```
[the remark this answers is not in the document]
harshula: Good question. We need to work out how to do CI on both the package's source repository and the package's spack file and then change this accordingly.

packages/oasis3-mct/package.py
harshula: # TODO: Clean up the debug / non-debug mechanism once the compiler
    #       flag audit is finished.
    #       https://github.com/ACCESS-NRI/ACCESS-OM/issues/12

packages/oasis3-mct/package.py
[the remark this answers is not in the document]
harshula: Had a chat with @echus on a plan. We have a package source ("package") and a Spack package definition ("spd"). We are proposing a "development" branch in both the package and spd repositories. e.g. We have a spd:development branch that depends on package:development. As changes from the "development" branch propagate to "main" we need to ensure that spd refers to package:main/master and not package:development.

harshula:
I'm going to drop the patch "oasis3-mct: add a nodebuginfo variant" for this PR and add a patch to point to "master" instead of "spack-build". I'm planning on using the variant name "deterministic" for builds that will be used to do a binary comparison.

harshula:
Ready for review!

packages/oasis3-mct/package.py
```diff
@@ -7,33 +7,37 @@
 
 from spack.package import install, join_path, mkdirp
 from pprint import pprint
+import os
```
aidanheerdegen, line 10: I don't think this import is used.

aidanheerdegen (approved):
Other than the import everything looks good to go.

packages/oasis3-mct/package.py
harshula: Thanks. I'll remove it.

packages/oasis3-mct/package.py
```diff
@@ -7,33 +7,37 @@
...
 import inspect
```
harshula, line 11: Removing "inspect" too.

PR #7, opened 2023-03-14
Update mom5 package.py

harshula:


packages/mom5/package.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def edit(self, spec, prefix):
+
+        srcdir = self.stage.source_path
+        makeinc_path = join_path(srcdir, "bin", "mkmf.template.spack")
+        config = {}
+
+        incs = (spec["oasis3-mct"].headers).cpp_flags + "/psmile.MPI1" + " "
+        for l in ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]:
+            incs += (spec[l].headers).cpp_flags + " "
```
aidanheerdegen, line 44:
```suggestion
        incs = [spec[l].headers).cpp_flags for l in ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]]
        incs = " ".join(incs)
```
Using a list comprehension is "pythonic" as is using a join to convert a list to a space separated string.

packages/mom5/package.py
```diff
@@ -0,0 +1,317 @@
...
+
+        # NOTE: The order of the libraries matter during the linking step!
+        # NOTE: datetime-fortran is a dependency of libaccessom2.
+        libs = "-L" + (spec["parallelio"].prefix) + "/lib -lpiof -lpioc "
```
aidanheerdegen, line 48:
```suggestion
        libs = ["-L" + join_path(spec["parallelio"].prefix, "lib"), "-lpiof", "-lpioc"]
```
As above, use a list ...

packages/mom5/package.py
```diff
@@ -0,0 +1,317 @@
...
+        for l in ["oasis3-mct", "libaccessom2", "netcdf-c", "netcdf-fortran", "datetime-fortran"]:
+            libs += (spec[l].libs).ld_flags + " "
```
aidanheerdegen, line 50:
```suggestion
        libs += [(spec[l].libs).ld_flags for l in ["oasis3-mct", "libaccessom2", "netcdf-c", "netcdf-fortran", "datetime-fortran"]]
        libs = " ".join(libs)
```
.. and add to it then convert to a space separated string.

packages/mom5/package.py
```diff
@@ -0,0 +1,317 @@
...
+        # cd ${ACCESS_OM_DIR}/src/mom/exp
+        # export mom_type=ACCESS-OM
+        # ./MOM_compile.csh --type $mom_type --platform spack
+
+        # cppDefs = "-Duse_netCDF -Duse_netCDF3 -Duse_libMPI -DUSE_OCEAN_BGC -DENABLE_ODA -DSPMD -DLAND_BND_TRACERS"
+        # set executable    = $root/exec/$platform/$type/fms_$type.x
+        # set srcList = ( accessom_coupler )
+        # set includes = "-I$executable:h:h/lib_FMS -I$executable:h:h/$type/lib_ocean"
+        # set libs = "$executable:h:h/$type/lib_ocean/lib_ocean.a $executable:h:h/lib_FMS/lib_FMS.a"
+        # bin/mkmf -f -m Makefile -a src -t bin/mkmfs.template.spack -p $executable:t" -o "$includes" -c "$cppDefs" -l "$libs"  $srcList
```
aidanheerdegen, line 301:
```suggestion
        # Build replicates the following shell command
        '''
        cd ${ACCESS_OM_DIR}/src/mom/exp
        export mom_type=ACCESS-OM
        ./MOM_compile.csh --type $mom_type --platform spack
        '''
```
I don't think the snippet from the `MOM_compile.csh` script is worth including.

packages/mom5/package.py
```diff
@@ -0,0 +1,317 @@
...
+    depends_on("oasis3-mct")
+    depends_on("datetime-fortran")
+    depends_on("netcdf-fortran@4.5.2:")
+    depends_on("netcdf-c@4.7.4:")
+    # TODO: For initial verification we are going to use static pio.
+    #       Eventually we plan to move to shared pio
+    # ~shared requires: https://github.com/spack/spack/pull/34837
+    depends_on("parallelio~pnetcdf~timing~shared")
+    depends_on("libaccessom2")
+    depends_on("hdf5")
```
aidanheerdegen, line 32: `cice5` doesn't have an explicit dependency on `hdf5` so I'm wondering if this is necessary?

packages/mom5/package.py
```diff
@@ -0,0 +1,317 @@
...
+# https://spack.readthedocs.io/en/latest/build_systems/makefilepackage.html
+class Mom5(MakefilePackage):
+    """MOM is a numerical ocean model based on the hydrostatic primitive equations."""
+
+    homepage = "https://www.access-nri.org.au"
+    git = "https://github.com/ACCESS-NRI/mom5.git"
+
+    maintainers = ["harshula"]
+
+    version("spack-build", branch="spack-build")
```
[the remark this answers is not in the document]
harshula: 1) Yes, I'll do the same as oasis3-mct PR.
2) The second question should be discussed in: https://github.com/ACCESS-NRI/ACCESS-OM/issues/6

packages/mom5/package.py
harshula: 1) In case you missed this:
```
MOM5.git/bin/environs.nci.gfortran
3:         module load hdf5
```
2) Since `netcdf-c` depends on `hdf5`, I'm happy to remove it, if you want.
3) Perhaps you could update `environs.nci.gfortran` too?

packages/mom5/package.py
```diff
@@ -0,0 +1,317 @@
...
+
+    phases = ["edit", "build", "install"]
+
+    def edit(self, spec, prefix):
+
+        srcdir = self.stage.source_path
+        makeinc_path = join_path(srcdir, "bin", "mkmf.template.spack")
+        config = {}
+
+        incs = (spec["oasis3-mct"].headers).cpp_flags + "/psmile.MPI1" + " "
```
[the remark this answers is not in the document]
harshula: 1) That change results in the following compile failure because you forgot to append a space at the end:
```
  >> 10134    Fatal Error: Cannot open module file 'mod_prism.mod' for reading 
              at (1): No such file or directory
```
2) Regardless, good catch! I should have used join_path() like I did three lines above it.

packages/mom5/package.py
harshula: 1) That change results in the following compile failure (string vs array):
`  >> 9916    make: *** [Makefile:45: fms_ACCESS-OM.x] Error 1`
2) Regardless, I should have used join_path().

packages/mom5/package.py
harshula: OK, happy to remove it. Please note that if we want to avoid using the `MOM_compile.csh`, that is what we'd have to do from this file.

packages/mom5/package.py
harshula: Does your suggestion traverse the list twice? Remember, we want to take a list and create a string in a single traversal of the list. List comprehension is great when the input and the desired output are lists.
Also, the code is more comprehensible the way it is.

packages/mom5/package.py
harshula: Does your suggestion traverse the list twice? Remember, we want to take a list and create a string in a single traversal of the list. List comprehension is great when the input and the desired output are lists.
Also, the code is more comprehensible the way it is.

harshula:
Ready for review!

packages/mom5/package.py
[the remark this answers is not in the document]
aidanheerdegen: Yes I should have said this was untested.

packages/mom5/package.py
aidanheerdegen: Yes I am suggesting traversing the list twice.

I think it is clearer to use two steps. It is just an opinion. If you disagree feel free to ignore this suggestion.

packages/mom5/package.py
aidanheerdegen: I should have said this was untested. Combined with the code below (which suggested using a `join`) I'd have thought it should work, assuming I didn't have a typo, so my apology if that wasted your time.

The thing I really don't think is good practice is appending spaces in the strings, e.g.
```
"-lpiof -lpioc "
```
they're easy to lose or not understand the importance of. 

in another PR you've done something like this:
```
"-lpiof -lpioc" + " "
```
which is preferable.

But ultimately this is why I suggested creating an array and then using `.join` as I think the intent is more explicit: the elements of the array are important tokens, the space is not. Once you put the space in the tokens you're imbuing it with more meaning than it has. Or to put it another way, you don't know if it is meaningful or not. IMO.

packages/mom5/package.py
aidanheerdegen: Yes I am suggesting traversing the list twice.

I think it is clearer to use two steps. It is just an opinion. If you disagree feel free to ignore this suggestion.

packages/mom5/package.py
aidanheerdegen: Yes I understood the intent of including it, but I thought it was confusing and a little misleading because that is just one bit of the file, so for example:
```
cppDefs = "-Duse_netCDF -Duse_netCDF3 -Duse_libMPI -DUSE_OCEAN_BGC -DENABLE_ODA -DSPMD -DLAND_BND_TRACERS"
```
is set as the default at the beginning (https://github.com/ACCESS-NRI/MOM5/blob/master/exp/MOM_compile.csh#L80) but later it is modified (https://github.com/ACCESS-NRI/MOM5/blob/master/exp/MOM_compile.csh#L93):
```
    set cppDefs  = ( "-Duse_netCDF -Duse_libMPI -DACCESS_OM" )
```
and again (https://github.com/ACCESS-NRI/MOM5/blob/master/exp/MOM_compile.csh#L117)
```
    set cppDefs = "$cppDefs -Duse_netCDF4"
```

Now I'm not arguing this is good design, but including the snippet just isn't that useful because of the convoluted logic in the build script.

packages/mom5/package.py
aidanheerdegen: I agree that it seems inconsistent with the `environs.nci.gfortran` but we **know** the spack `netcdf-c` package depends on it already, and in a non-straightforward way, so best to leave `netcdf-c` to specify that

https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/netcdf-c/package.py#L128-L142

Removing it from `environs.nci.gfortran` should be a separate issue.

packages/mom5/package.py
```diff
@@ -0,0 +1,320 @@
...
+    _platform = "spack"
+    _mom_type = "ACCESS-OM"
+
+    def edit(self, spec, prefix):
+
+        srcdir = self.stage.source_path
+        makeinc_path = join_path(srcdir, "bin", "mkmf.template.spack")
+        config = {}
+
+        incs = join_path((spec["oasis3-mct"].headers).cpp_flags, "psmile.MPI1 ")
```
aidanheerdegen, line 44: As expressed in a previous comment, I don't think trailing spaces in strings is a good approach and prefer what you have done in other packages and below in the loop:
```suggestion
        incs = join_path((spec["oasis3-mct"].headers).cpp_flags, "psmile.MPI1") + " "
```

packages/mom5/package.py
aidanheerdegen: Note: to be specific the second traversal isn't the same list, the `psmile.MPI1` include at the beginning and the various `cpp_flags` lookups have taken place for the listed dependencies. And yes, I am suggesting traversing **that** list.

packages/mom5/package.py
```diff
@@ -0,0 +1,320 @@
...
+        for l in ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]:
+            incs += (spec[l].headers).cpp_flags + " "
+
+        # NOTE: The order of the libraries matter during the linking step!
+        # NOTE: datetime-fortran is a dependency of libaccessom2.
+        libs = ("-L" + join_path(spec["parallelio"].prefix, "lib ")
+                + "-lpiof -lpioc ")
```
aidanheerdegen, line 51: As expressed in a previous comment, I don't think trailing spaces in strings is a good approach and prefer what you have done in other packages and below in the loop:
```suggestion
        libs = ("-L" + join_path(spec["parallelio"].prefix, "lib") + " "
                + "-lpiof -lpioc" + " ")
```

aidanheerdegen (approved):
Happy for you to incorporate what you think is useful.

packages/mom5/package.py
harshula: Yes, that is clearer.

packages/mom5/package.py
harshula: Yes, that is clearer.

packages/mom5/package.py
harshula:
```
>>> import timeit
>>> h = """\
... lut = {"parallelio": "a", "oasis3-mct": "b", "libaccessom2": "c", "netcdf-fortran": "d"}
... incs = ""
... for l in ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]:
...     incs += lut[l] + " "
... """
>>> a = """\
... lut = {"parallelio": "a", "oasis3-mct": "b", "libaccessom2": "c", "netcdf-fortran": "d"}
... incs = [lut[l] for l in ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]]
... inc = " ".join(incs)
... """
>>> timeit.timeit(stmt=h, number=10000000)
2.131908037001267
>>> timeit.timeit(stmt=a, number=10000000)
2.251114501967095
>>> timeit.timeit(stmt=h, number=10000000)
2.135053211997729
>>> timeit.timeit(stmt=a, number=10000000)
2.2056077060406096
>>> timeit.timeit(stmt=h, number=10000000)
2.1651839180267416
>>> timeit.timeit(stmt=a, number=10000000)
2.2446520270314068
```

packages/mom5/package.py
harshula: 1) Performance difference:
```
>>> import timeit
>>> h = """\
... lut = {"parallelio": "a", "oasis3-mct": "b", "libaccessom2": "c", "netcdf-fortran": "d"}
... incs = ""
... for l in ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]:
...     incs += lut[l] + " "
... """
>>> a = """\
... lut = {"parallelio": "a", "oasis3-mct": "b", "libaccessom2": "c", "netcdf-fortran": "d"}
... incs = [lut[l] for l in ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]]
... inc = " ".join(incs)
... """
>>> timeit.timeit(stmt=h, number=10000000)
2.131908037001267
>>> timeit.timeit(stmt=a, number=10000000)
2.251114501967095
>>> timeit.timeit(stmt=h, number=10000000)
2.135053211997729
>>> timeit.timeit(stmt=a, number=10000000)
2.2056077060406096
>>> timeit.timeit(stmt=h, number=10000000)
2.1651839180267416
>>> timeit.timeit(stmt=a, number=10000000)
2.2446520270314068
```
2) The suggested code is not clearer. However, there may be a better way to do it than both our approaches.

packages/mom5/package.py
aidanheerdegen: I don't think performance is as important as clarity. I don't necessarily think the list comprehension is much clearer than the loop, simply that once you're just doing a lookup and there is no other logic it might be easier to use a comprehension, but I think this is perfectly fine:
```python
lut = {"parallelio": "a", "oasis3-mct": "b", "libaccessom2": "c", "netcdf-fortran": "d"}
incs = []
for l in ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]:
     incs += lut[l]
inc = " ".join(incs)
```

Agreed there are other methods: map, formatted string, class with a method
```python
>>> " ".join(map(lambda l: lut[l], ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]))
'a b c d'
```
but I don't think `map` is necessarily clearer.

packages/mom5/package.py
harshula: We've conflated two issues:
1) Constructing a string from a strings.
2) Constructing a string from a list.

For (1), yes join() appears to be a better solution.
For (2), join() appears to traverse the list twice and the suggested approach is too frankensteinish for my liking.

aidanheerdegen: Heh. I'd suggest replacing "frankensteinish" with "pythonic". It is a common pattern in python, and so the intent is recognisable.

As I said I really don't mind except I don't think we should have strings with magic spaces at the end that break the functionality if the space is removed. On that we appear to be in agreement, so the rest is up to you. Seriously, do what you makes most sense to you.

packages/mom5/package.py
harshula: This avoids appending a space, avoids traversing the list twice and uses join():
```
        istr = join_path((spec["oasis3-mct"].headers).cpp_flags, "psmile.MPI1")
        ideps = ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]
        incs = " ".join([istr] + [(spec[d].headers).cpp_flags for d in ideps])
```

packages/mom5/package.py
harshula: This avoids appending a space, avoids traversing the list twice and uses join():
```
        ldeps = ["oasis3-mct", "libaccessom2", "netcdf-c", "netcdf-fortran", "datetime-fortran"]
        libs = " ".join([lstr] + [(spec[d].libs).ld_flags for d in ldeps])
```

packages/mom5/package.py
harshula: Here's an unfrankensteinish pythonic approach:
```
        lstr = " ".join(
                    ["".join("-L", join_path(spec["parallelio"].prefix, "lib")),
                    "-lpiof",
                    "-lpioc"]
                   )
        ldeps = ["oasis3-mct", "libaccessom2", "netcdf-c", "netcdf-fortran", "datetime-fortran"]
        libs = " ".join([lstr] + [(spec[d].libs).ld_flags for d in ldeps])
```

packages/mom5/package.py
aidanheerdegen: Indeed. Looks clean and clear.

packages/mom5/package.py
aidanheerdegen: I like it. Very clean and clear.

packages/mom5/package.py
harshula: What I ended up writing:
```
        lstr = " ".join(
                    ["".join(["-L", join_path(spec["parallelio"].prefix, "lib")]),
                    "-lpiof",
                    "-lpioc"]
                   )
```

PR #9, opened 2023-03-14
Update datetime-fortran package.py

harshula:


packages/datetime-fortran/package.py
```diff
@@ -39,3 +39,13 @@ class DatetimeFortran(AutotoolsPackage):
     version('1.4.3', sha256='e3cb874fde0d36a97e282689eef459e8ce1973183d9acd95568149bc2d74436d')
     version('1.4.2', sha256='5b70c6e5d38032951e879b437e9ac7c5d483860ce8a9f6bbe6f1d6cd777e737f')
     version('1.4.1', sha256='4a178b63301f0016b7634625062278742a44026f4c37cafcb9e8ba9649db85e0')
+
+    patch("0001-Enable-deterministic-builds-using-D-flag-for-ar.patch", when="@1.7.0")
```
aidanheerdegen, line 43: Just checking, do you want this patch to apply on when version is only `==1.7.0` or `>=1.7.0`?

harshula: Good question. It was only tested with 1.7.0. I'll leave it to the maintainer to test other versions. ;-)

aidanheerdegen (approved):
Good to go. I think I'll merge.

harshula:
For small merges like this, let's do merge with rebase. Merge with commit is just unnecessary noise.

PR #16, opened 2023-05-12
mom5: parallelio is not a dependency

### LINKED ISSUE #15: mom5: does mom5 require parallelio?
Based on bin/mkmf.template.nci:
```
-I$(OASIS_ROOT)/Linux/build/lib/pio \
```
I added parallelio as a dependency in the mom5 SPD. However, I don't see any calls to pio_* functions in the source code. It also builds without parallelio as a dependency.

harshula:
Close #15

### DIFF
--- a/packages/mom5/package.py
+++ b/packages/mom5/package.py
@@ -27,10 +27,6 @@ class Mom5(MakefilePackage):
     depends_on("datetime-fortran")
     depends_on("netcdf-fortran@4.5.2:")
     depends_on("netcdf-c@4.7.4:")
-    # TODO: For initial verification we are going to use static pio.
-    #       Eventually we plan to move to shared pio
-    # ~shared requires: https://github.com/spack/spack/pull/34837
-    depends_on("parallelio~pnetcdf~timing~shared")
     depends_on("libaccessom2")
 
     phases = ["edit", "build", "install"]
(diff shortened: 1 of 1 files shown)


packages/mom5/package.py
```diff
@@ -45,18 +41,13 @@ def edit(self, spec, prefix):
         config = {}
 
         istr = join_path((spec["oasis3-mct"].headers).cpp_flags, "psmile.MPI1")
-        ideps = ["parallelio", "oasis3-mct", "libaccessom2", "netcdf-fortran"]
+        ideps = ["oasis3-mct", "libaccessom2", "netcdf-fortran"]
```
aidanheerdegen, line 44: I'm sure it isn't an error because this isn't a change and it compiled previously, but I am curious why an include path for `netcdf-c` isn't required here. I was fairly sure FMS (part of MOM) has calls to the C layer of netcdf, which would require a header file I'd guess.

harshula: Good question! Not sure how, but netcdf-c-4.7.4-jzburrwbsqhi6cjth7nm4zdnunhcyhaq/include appears in the -I argument to gcc.

```
grep ^gcc spack-upstream.git/opt/spack/linux-debian12-x86_64/gcc-12.2.0/mom5-master-r2wkz46epnpfiexqjvhigmuu76knxfbh/.spack/spack-build-out.txt | wc -l
9

$ grep ^gcc spack-upstream.git/opt/spack/linux-debian12-x86_64/gcc-12.2.0/mom5-master-r2wkz46epnpfiexqjvhigmuu76knxfbh/.spack/spack-build-out.txt | grep netcdf-c | wc -l
9

$ grep ^gcc spack-upstream.git/opt/spack/linux-debian12-x86_64/gcc-12.2.0/mom5-master-r2wkz46epnpfiexqjvhigmuu76knxfbh/.spack/spack-build-out.txt | grep netcdf-fortran | wc -l
9
```

harshula: @aidanheerdegen In this situation, when a question is *not* germane to the code change, it would be better to focus on the code review and open a separate issue to pursue tangential questions.

harshula: GCC compiles contain the Makefile snippet:
```
CFLAGS += $(shell nc-config --cflags)
```
that results in the following being added: 
```
-I$SPACK/opt/spack/linux-debian12-x86_64/gcc-12.2.0/netcdf-c-4.7.4-jzburrwbsqhi6cjth7nm4zdnunhcyhaq/include -I$SPACK/opt/spack/linux-debian12-x86_64/gcc-12.2.0/hdf5-1.12.2-igvxfglba4p6pcufko2mbprzeampoljh/include
```

harshula: Opened #17

aidanheerdegen (approved):
LGTM

aidanheerdegen (approved):
LGTM

PR #18, opened 2023-05-15
access-om2: create a new bundle package

harshula:


packages/access-om2/package.py
```diff
@@ -0,0 +1,23 @@
...
+from spack.package import *
+
+class AccessOm2(BundlePackage):
+    """ACCESS-OM2 bundle contains the coupled CICE5 and MOM5 models."""
+
+    homepage = "https://www.access-nri.org.au"
+
+    maintainers = ["harshula"]
+
+    version("master")
```
aidanheerdegen, line 17: GitHub has transitioned to using `main` as the name of the default branch, and this repo is using that convention. To be consistent we should use `main` as the version if the intention that it should use the current `main` branch.

aidanheerdegen (approved):
Happy to merge with or without that change, up to you.

packages/access-om2/package.py
harshula: Good catch! In a bundle (SBD) `version` has a different meaning to an SPD. In an SPD, it's the version of the source package. In this case, I assume it's the version of the SBD. So should we use a date based version? e.g. "202305-1"

aidanheerdegen: Agreed calendar based versioning best, the [dev-docs suggest](https://github.com/ACCESS-NRI/dev-docs/wiki/Versioning) `2023.05.001`

PR #26, opened 2023-07-13
Workflow For Notifying ACCESS-NRI/build-ci that spack_packages is Updated

CodeGat:
GitHub Actions can be strange. In order for build-ci to know that spack_packages has been updated, spack_packages needs to notify build-ci that there has been a change. In this PR, workflows are added to notify build-ci that spack_packages has had a push to master, or a tag released. 

Related PR from build-ci: https://github.com/ACCESS-NRI/build-ci/pull/19

Should go partway to closing ACCESS-NRI/build-ci#4 !

### DIFF
--- a/.github/workflows/dispatch-update-to-build-ci.yml
+++ b/.github/workflows/dispatch-update-to-build-ci.yml
@@ -6,12 +6,24 @@ on:
     tags:
       - "*.*.*"
 jobs:
+  extract-tag:
+    runs-on: ubuntu-latest
+    outputs:
+      tag: ${{ steps.tag-step.outputs.tag }}
+    steps:
+    - name: Clone repo
+      uses: actions/checkout@v3
+    - name: Extract tag
+      id: tag-step
+      run: echo "tag=${GITHUB_REF#refs/*/}" >> $GITHUB_OUTPUT
+      
   send-update-to-build-ci:
     name: Send Update to Build CI repo
     runs-on: ubuntu-latest
+    needs: extract-tag
     steps:
       - name: Send event to build-ci repo
         run: |
-          curl -XPOST -u "${{ secrets.RELEASE_BUILD_CI_PAT_USERNAME }}:${{ secrets.RELEASE_BUILD_CI_PAT }}" -H "Accept: application/vnd.github.everest-preview+json" -H "Content-Type: application/json" https://api.github.com/repos/ACCESS-NRI/build-ci/actions/workflows/build-and-push-image-base-spack.yml/dispatches --data '{"ref": "main"}'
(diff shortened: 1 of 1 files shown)


.github/workflows/dispatch-update-to-build-ci.yml
```diff
@@ -0,0 +1,48 @@
+name: "Dispatch Event to Build CI"
+on:
+  push:
+    branches:
+      - main
+    tags: 
+      - '*.*.*'
+    paths:
+      - '**.py'
```
Quartzer2, line 9: Possibly need to check if files other than.py are changed?

For example packages/cice5/spack-build.sh

.github/workflows/dispatch-update-to-build-ci.yml
```diff
@@ -0,0 +1,48 @@
...
+  send-branch-update-to-build-ci:
+    name: Send Branch Update to Build CI repo
+    runs-on: ubuntu-latest
+    if: github.ref_type == 'branch'
+    steps:
+      - name: Get names of directories modified in push/tag
+        id: directories
+        run: |
+          # get all the changed files in the push, only keep the parent directory (corresponds to the package name), and turn it into a space-separated list!
+          echo "modified=$(git diff --name-only ${{ github.event.after }} ${{ github.event.before }} | cut -d '/' -f 2 | tr '\n' ' ')" >> $GITHUB_OUTPUT
```
Quartzer2, line 41: Should this snippet be a stand-alone shell script that we could test? Possible a separate shelll script within the repository that could be changed depending on circumstances.

Better to have a test for the script itself than diagnosing down the track?

Quartzer2 (approved):
Looks good. Just a couple of things to look at.

.github/workflows/dispatch-update-to-build-ci.yml
CodeGat: Noted and changed :)

.github/workflows/dispatch-update-to-build-ci.yml
CodeGat: Good point. Although, are Github contexts (`${{ github.whatever }}`) available from within the script?
From a cursory glance around, it's not accessible, so people pass in the context as env vars instead. But, we have secrets in that curl request (`${{ secrets.RELEASE_BUILD_CI_PAT(_USERNAME) }}`) which could get exposed through passing them as env vars. I'm leaning against it if this is the case.

Quartzer2: True. Maybe for further dicussion down the track.

Quartzer2 (approved):

.github/workflows/dispatch-update-to-build-ci.yml
aidanheerdegen: Could also change to
```yaml
    paths:
      - 'packages/**'
```
to avoid triggering if README or `repo.yaml` is changed.

CodeGat: Good point!! Will add now

.github/workflows/dispatch-update-to-build-ci.yml
aidanheerdegen: I think the stand-alone snippet suggestion just applied to the shell magic to extract the modified directories. This might involve making a script like this (named `./util/find_modified_packages.sh`):
```
#!/bin/bash
git diff --name-only $1 $2 | cut -d '/' -f 2 | tr '\n' ' '
```
and changing the CI command to
```suggestion
          echo "modified=$(./util/find_modified_packages.sh ${{ github.event.after }} ${{ github.event.before }} )" >> $GITHUB_OUTPUT
```
and add CI to test `./util/find_modified_packages.sh`

PR #31, opened 2023-08-02
cice5: explicitly set the rpath of library dependencies

### LINKED ISSUE #14: cice5: shared libraries not found
```
$ ldd opt/spack/linux-rocky8-x86_64/intel-2019.5.281/cice5-master-byagur66qd4y7za7vauskh5w34t3cb54/bin/cice_auscom_1440x1080_48x40_480p.exe | grep not
	libnetcdf.so.18 => not found
	libnetcdff.so.7 => not found
```

harshula:
* Print the generated config file to stdout.
* Close #14.

### DIFF
--- a/packages/cice5/package.py
+++ b/packages/cice5/package.py
@@ -63,6 +63,24 @@ class Cice5(MakefilePackage):
     def url_for_version(self, version):
         return "https://github.com/ACCESS-NRI/cice5/tarball/{0}".format(version)
 
+    # The reason for the explicit -rpath is:
+    # https://github.com/ACCESS-NRI/spack_packages/issues/14#issuecomment-1653651447
+    def get_linker_args(self, spec, name):
+        return " ".join(
+                    [(spec[name].libs).ld_flags,
+                    "-Wl,-rpath=" + join_path(spec[name].prefix, "lib")]
+                   )
+
+    # The reason for the explicit -rpath is:
+    # https://github.com/ACCESS-NRI/spack_packages/issues/14#issuecomment-1653651447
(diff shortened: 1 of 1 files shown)


packages/cice5/package.py
```diff
@@ -63,6 +63,24 @@ class Cice5(MakefilePackage):
...
+    def get_linker_args(self, spec, name):
+        return " ".join(
+                    [(spec[name].libs).ld_flags,
+                    "-Wl,-rpath=" + join_path(spec[name].prefix, "lib")]
+                   )
+
+    # The reason for the explicit -rpath is:
+    # https://github.com/ACCESS-NRI/spack_packages/issues/14#issuecomment-1653651447
+    def make_linker_args(self, spec, name, namespecs):
+        path = join_path(spec["parallelio"].prefix, "lib")
```
aidanheerdegen, line 77: Should this be
```suggestion
        path = join_path(spec[name].prefix, "lib")
```

aidanheerdegen (changes requested):
I think just that one change required.

packages/cice5/package.py
harshula: Yes, that's a bug! Good catch.

aidanheerdegen (approved):
Looks good to me. If you had the output of the linker flags handy in a buffer somewhere it would be nice to paste it into the PR to show how all that logic pans out in an example build, but don't bother running it just to get that.

harshula:
Good idea. I have it in my notes, so I've added it to the issue where the rest of the documentation resides.

PR #33, opened 2023-08-04
CI refers to the build-and-push-image-build.yml workflow

CodeGat:
Since all of the packages will be part of the same image, there will be no need to conditionally compile some of them. We'll just use the regular workflow that compiles all of them. 

Should go partway to closing https://github.com/ACCESS-NRI/build-ci/issues/9 !

### DIFF
--- a/.github/workflows/dispatch-update-to-build-ci.yml
+++ b/.github/workflows/dispatch-update-to-build-ci.yml
@@ -14,20 +14,12 @@ jobs:
     runs-on: ubuntu-latest
     if: github.ref_type == 'tag'
     steps:
-      - uses: actions/checkout@v3
-      
-      - name: Get names of all package directories
-        id: directories
-        run: |
-          # list all the packages, turn it into a space-separated list!
-          echo "modified=$(ls packages | tr '\n' ' ')" >> $GITHUB_OUTPUT
-
       - name: Send event and tag to build-ci repo
         run: |
           curl -XPOST -u "${{ secrets.RELEASE_BUILD_CI_PAT_USERNAME }}:${{ secrets.RELEASE_BUILD_CI_PAT }}" \
             -H "Accept: application/vnd.github.everest-preview+json" \
             -H "Content-Type: application/json" https://api.github.com/repos/ACCESS-NRI/build-ci/actions/workflows/build-and-push-image-build.yml/dispatches \
-            --data '{"ref": "main", "inputs":{"spack-packages-version": "${{ github.ref_name }}"}}'
+            --data '{"ref": "main", "inputs":{"spack-packages-build-required": "true", "spack-packages-version": "${{ github.ref_name }}", "model": "all"}}'
 
   send-branch-update-to-build-ci:
     name: Send Branch Update to Build CI repo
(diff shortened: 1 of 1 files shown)


.github/workflows/dispatch-update-to-build-ci.yml
```diff
@@ -40,9 +32,9 @@ jobs:
           # get all the changed files in the push, only keep the parent directory (corresponds to the package name), and turn it into a space-separated list!
           echo "modified=$(git diff --name-only ${{ github.event.after }} ${{ github.event.before }} | cut -d '/' -f 2 | tr '\n' ' ')" >> $GITHUB_OUTPUT
```
aidanheerdegen, line 33: Do we still need this? It doesn't seem to be used.

.github/workflows/dispatch-update-to-build-ci.yml
```diff
@@ -14,20 +14,12 @@ jobs:
...
-          echo "modified=$(ls packages | tr '\n' ' ')" >> $GITHUB_OUTPUT
-
       - name: Send event and tag to build-ci repo
         run: |
           curl -XPOST -u "${{ secrets.RELEASE_BUILD_CI_PAT_USERNAME }}:${{ secrets.RELEASE_BUILD_CI_PAT }}" \
             -H "Accept: application/vnd.github.everest-preview+json" \
-            -H "Content-Type: application/json" https://api.github.com/repos/ACCESS-NRI/build-ci/actions/workflows/notified-of-spack-packages-update.yml/dispatches \
-            --data '{"ref": "main", "inputs":{"spack-packages-version": "${{ github.ref_name }}", "packages": "${{ steps.directories.outputs.modified }}"}}'
+            -H "Content-Type: application/json" https://api.github.com/repos/ACCESS-NRI/build-ci/actions/workflows/build-and-push-image-build.yml/dispatches \
+            --data '{"ref": "main", "inputs":{"spack-packages-build-required": "true", "spack-packages-version": "${{ github.ref_name }}", "model": "all"}}'
```
aidanheerdegen, line 22: I wonder if we need to specify `"model": "all"` here. I'm normally a fan of being specific about things and not relying on defaults, but it seems to be in this case it might be sufficient to let the CI decide what it wants to build by default.

I can imagine a time when we no longer build a model routinely, and we might not want `all` here. Of course `all` could be redefined in the CI repo.

So I'm not super-worried about it either way, just wanted to make sure the options had been canvassed.

.github/workflows/dispatch-update-to-build-ci.yml
```diff
@@ -40,9 +32,9 @@ jobs:
...
           
-      - name: Send event and tag to build-ci repo
+      - name: Send event and branch to build-ci repo
         run: |
           curl -XPOST -u "${{ secrets.RELEASE_BUILD_CI_PAT_USERNAME }}:${{ secrets.RELEASE_BUILD_CI_PAT }}" \
             -H "Accept: application/vnd.github.everest-preview+json" \
-            -H "Content-Type: application/json" https://api.github.com/repos/ACCESS-NRI/build-ci/actions/workflows/notified-of-spack-packages-update.yml/dispatches \
-            --data '{"ref": "main", "inputs":{"spack-packages-version": "${{ github.ref_name }}", "packages": "${{ steps.directories.outputs.modified }}"}}'
+            -H "Content-Type: application/json" https://api.github.com/repos/ACCESS-NRI/build-ci/actions/workflows/build-and-push-image-build.yml/dispatches \
+            --data '{"ref": "main", "inputs":{"spack-packages-build-required": "true", "spack-packages-version": "${{ github.ref_name }}", "model": "all"}}'
```
aidanheerdegen, line 40: See above re `"model": "all"`.

.github/workflows/dispatch-update-to-build-ci.yml
aidanheerdegen: Also, do we want to run this for all branches? Is that what this is doing?

aidanheerdegen (approved):
Some questions, as per usual.

.github/workflows/dispatch-update-to-build-ci.yml
CodeGat: Thats fair enough. I'm going to be thinking of a way to determine which is access-om2 vs access-om3 based on the directories pushed. Might ask @harshula about ways to do this programmatically based on the things in the `package.py`s so we don't have to update a file or something every time a new `package.py` is added.

.github/workflows/dispatch-update-to-build-ci.yml
CodeGat: No, it'll only run for the `main` branch. And yes, it won't be used it we remove the ability to created packages based on pushes to the branch name, although I can't think of a way to determine which models to be rebuilt based on tag.

.github/workflows/dispatch-update-to-build-ci.yml
CodeGat: Resolved offline with @aidanheerdegen and @harshula - Will rebuild every time, as we need to be able to clone a tag and have it all there. I'll raise a ticket for it as an investigation, however.

.github/workflows/dispatch-update-to-build-ci.yml
CodeGat: See https://github.com/ACCESS-NRI/spack_packages/pull/33#discussion_r1295230111

PR #34, opened 2023-08-04
Add a deterministic variant

harshula:
* This PR is only focused on adding the deterministic variant.
* The general compiler flag audit will happen in https://github.com/ACCESS-NRI/ACCESS-OM/issues/12
* We'll likely end-up storing the compiler flags in an appropriate data structure and then extract the relevant compiler flags based on the variant/compiler.

### DIFF
--- a/packages/cice5/package.py
+++ b/packages/cice5/package.py
@@ -18,17 +18,21 @@ class Cice5(MakefilePackage):
 
     version("master", branch="master")
 
+    variant("deterministic", default=False, description="Deterministic build.")
+
     # Depend on virtual package "mpi".
     depends_on("mpi")
-    depends_on("oasis3-mct")
-    depends_on("datetime-fortran")
     depends_on("netcdf-fortran@4.5.2:")
     depends_on("netcdf-c@4.7.4:")
     # TODO: For initial verification we are going to use static pio.
     #       Eventually we plan to move to shared pio
     # ~shared requires: https://github.com/spack/spack/pull/34837
     depends_on("parallelio~pnetcdf~timing~shared")
-    depends_on("libaccessom2")
+    depends_on("datetime-fortran")
+    depends_on("oasis3-mct+deterministic", when="+deterministic")
+    depends_on("oasis3-mct~deterministic", when="~deterministic")
+    depends_on("libaccessom2+deterministic", when="+deterministic")
+    depends_on("libaccessom2~deterministic", when="~deterministic")
 
     phases = ["edit", "build", "install"]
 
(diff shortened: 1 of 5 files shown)


packages/cice5/package.py
```diff
@@ -101,6 +105,13 @@ def edit(self, spec, prefix):
         ldeps = ["oasis3-mct", "libaccessom2", "netcdf-c", "netcdf-fortran", "datetime-fortran"]
         libs = " ".join([lstr] + [self.get_linker_args(spec, d) for d in ldeps])
 
+        # TODO: https://github.com/ACCESS-NRI/ACCESS-OM/issues/12
+        NCI_OPTIM_FLAGS = "-g3 -O2 -axCORE-AVX2 -debug all -check none -traceback -qopt-report=5 -qopt-report-annotate -assume buffered_io"
+        CFLAGS = "-c -O2"
+        if "+deterministic" in self.spec:
+            NCI_OPTIM_FLAGS = "-g0 -O0 -axCORE-AVX2 -debug none -check none -qopt-report=5 -qopt-report-annotate -assume buffered_io"
```
aidanheerdegen, line 112: Deterministic builds require `-O0`?

harshula: Without the `-O0` , it may default to an undesirable optimisation level:
```
ifort(1):

       -O[n]

              Specifies the code optimization for
                     applications.

              Arguments:

              n                 Is the optimization level. Possible values are
                                1,  2, or 3. On Linux* and macOS* systems, you
                                can also specify 0.

              Default:

              O2                Optimizes for code  speed.  This  default  may
                                change   depending  on  which  other  compiler
                                options are specified. For details, see below.
```

aidanheerdegen (approved):
Looks good.

packages/cice5/package.py
aidanheerdegen: Just a follow-up: it is desirable for the reproducible tests to be the production builds, or as close as possible. Then we can use build reproducibility as a proxy for model replicability/reproducibility.

harshula: I had a chat to Aidan, his actual question was, "Is `-O2` non-deterministic?". I'll check.

micaeljtoliveira: >  "Is -O2 non-deterministic?" 

@harshula Yes, it is. You can still use -O2, but then you need to specify the floating point model to make it deterministic. See the [Intel compiler manual](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-2/fp-model-fp.html#GUID-63514789-6BE7-4FA8-9389-F3B618A2EE1A). You might also want to set ` -qopt-dynamic-align` (see [here](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-2/qopt-dynamic-align-qopt-dynamic-align.html)).

PR #41, opened 2023-08-30
libaccessom2: allow variants to directly modify CMakeLists.txt

harshula:
Modify CMakeLists.txt directly from the SPD instead of having a two step process, where we have to add the variant to the SPD and then go modify the package's source code repository.

### DIFF
--- a/packages/libaccessom2/package.py
+++ b/packages/libaccessom2/package.py
@@ -38,6 +38,14 @@ class Libaccessom2(CMakePackage):
     def url_for_version(self, version):
         return "https://github.com/ACCESS-NRI/libaccessom2/tarball/{0}".format(version)
 
-    def cmake_args(self):
-        return [self.define_from_variant('DETERMINISTIC', "deterministic"),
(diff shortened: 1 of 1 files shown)


packages/libaccessom2/package.py
```diff
@@ -38,6 +38,14 @@ class Libaccessom2(CMakePackage):
     def url_for_version(self, version):
         return "https://github.com/ACCESS-NRI/libaccessom2/tarball/{0}".format(version)
 
-    def cmake_args(self):
```
CodeGat, line 41: Just a quick check: did anything call this function that would miss it being deleted?

harshula: My understanding is that is a function over-ride mechanism as described in https://spack.readthedocs.io/en/latest/build_systems/cmakepackage.html#adding-flags-to-cmake .

CodeGat: Noted.

CodeGat (approved):
Talked about this over a call. Approved.

Issue #20, opened 2023-06-29
issm: install/benchmarking on gadi


fmccormack:
I'm after an install of the Ice-sheet and Sea-level System Model (ISSM) on gadi. 

ISSM is open source. Standard install requires autotools, cmake, and petsc, as well as triangle, chaco and m1qn3. Basic install instructions can be found in the user guide (https://issm.jpl.nasa.gov/files/userguide.pdf), and examples of configure scripts for different clusters can be found on the wiki (https://issm.ess.uci.edu/trac/issm/wiki).  We wouldn't use the python or matlab wrappers (with-wrappers=no in configure script) when running on gadi. 

I've had ISSM running in the past on gadi, but with more recent versions of ISSM I've had issues. A recent version of ISSM didn't compile with the system petsc modules, so I installed a local version of petsc. With this configuration the job would run, but occasionally no output would be written and there were no error messages. My postdoc Johanna was able to get the most recent version of ISSM installed with a system module of petsc (see attached "install_issm" script; you'll also need to export the variable $ISSM_DIR to be the install directory), but I haven't tested this build thoroughly. I'd be keen for advice on appropriate optimisation flags with this.

If the Underworld folks have advice on petsc builds and scaling that'd be great.

There are many different solver options for running ISSM (see examples in the attached paper), which are configured at the user end. Once we get an optimised build, I'd be very keen to perform benchmarking tests for parallel performance.

[install_issm.txt](https://github.com/ACCESS-NRI/spack_packages/files/11900748/install_issm.txt)
[environment.txt](https://github.com/ACCESS-NRI/spack_packages/files/11900749/environment.txt)
[gmd-10-155-2017.pdf](https://github.com/ACCESS-NRI/spack_packages/files/11900753/gmd-10-155-2017.pdf)

aidanheerdegen:
For linking purposes, this started as a discussion on the ACCESS-Hive Forum

https://forum.access-hive.org.au/t/ice-sheet-model-install-and-benchmarking-on-gadi/912

aidanheerdegen:
The result of 5 minutes of googling:

The Parallel Ice Sheet Model (`PISM`) has a spack package and also relies on `PETSc`:

https://www.pism.io/docs/installation/spack.html

[Firedrake](https://github.com/firedrakeproject/firedrake) has a spack package

https://github.com/firedrakeproject/firedrake-spack

and also has `PETSc` as a dependency. @angus-g have you tried a firedrake spack install?

[...]

PR #42, opened 2023-09-12
Add the ISSM model and supporting packages

angus-g:
As mentioned in #20, here is the ISSM package which pulls in PETSc and builds using Intel compilers and MKL on Gadi. The `packages.yaml` may need to be modified for site configurations to specify that `intel-oneapi-mkl` uses the `+cluster` variant in order to include ScaLAPACK and BLACS.

It's also possible that `intel-oneapi-mkl` can be renamed to something more NCI-specific, like the existing `nci-openmpi` package.

### DIFF
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -1,0 +1,53 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# Copyright 2023 Angus Gibson
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Issm(AutotoolsPackage):
+    """Ice-sheet and Sea-Level System Model"""
+
+    homepage = "https://issm.jpl.nasa.gov/"
+    svn = "https://issm.ess.uci.edu/svn/issm/issm/trunk"
+
+    version("develop")
+
+    depends_on("autoconf", type="build")
+    depends_on("automake", type="build")
+    depends_on("libtool", type="build")
+    depends_on("m4", type="build")
+
+    depends_on("mpi")
+    depends_on("petsc+metis+mumps+scalapack")
+    depends_on("m1qn3")
+
+    def autoreconf(self, spec, prefix):
+        autoreconf("--install", "--verbose", "--force")
+
+    def configure_args(self):
+        args = [
+          "--with-wrappers=no",
+          "--enable-debugging",
+          "--enable-development",
+          "--enable-shared",
+          "--without-kriging",
+        ]
+        args.append("--with-petsc-dir={0}".format(self.spec["petsc"].prefix))
+        args.append("--with-metis-dir={0}".format(self.spec["metis"].prefix))
+        args.append("--with-mumps-dir={0}".format(self.spec["mumps"].prefix))
+        args.append("--with-m1qn3-dir={0}".format(self.spec["m1qn3"].prefix.lib))
+
+        # Even though we set the MPI compilers manually, the build system
+        # wants us to explicitly request an MPI-enabled build by telling
+        # it the MPI include directory.
+        args.append("--with-mpi-include={0}".format(self.spec["mpi"].prefix.include))
+        args.append("CC=" + self.spec["mpi"].mpicc)
+        args.append("CXX=" + self.spec["mpi"].mpicxx)
+        args.append("FC=" + self.spec["mpi"].mpifc)
+        args.append("F77=" + self.spec["mpi"].mpif77)
+
+        return args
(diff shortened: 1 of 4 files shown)


aidanheerdegen:
This is great @angus-g, thanks.

I agree it would be good for the package to be called `nci-intel-oneapi-mkl` to be clear what it does.

Do you imagine that you'll make a PR for `m1qn3` and `triangle` to the main spack repo?

I'll try and give the build a burl tomorrow.

angus-g:
> Do you imagine that you'll make a PR for m1qn3 and triangle to the main spack repo?

I'll have to verify the license for m1qn3, but I could do that, yes.

harshula:
Hi @aidanheerdegen , @angus-g is coming over to the office on Tuesday afternoon to walk-through the PR.

harshula:
We did an in-person review today. TODO: Licenses, header comment in patch file, do `from spack.package import install, mkdirp` instead of `from spack.package import *`, etc.

packages/issm/package.py
```diff
@@ -0,0 +1,50 @@
...
+        args.append("--with-petsc-dir={0}".format(self.spec["petsc"].prefix))
+        args.append("--with-metis-dir={0}".format(self.spec["metis"].prefix))
+        args.append("--with-mumps-dir={0}".format(self.spec["mumps"].prefix))
+        args.append("--with-m1qn3-dir={0}".format(self.spec["m1qn3"].prefix.lib))
+        args.append("--with-mpi-include={0}".format(self.spec["mpi"].prefix.include))
+
+        args.append("CC=" + self.spec["mpi"].mpicc)
+        args.append("CXX=" + self.spec["mpi"].mpicxx)
+        args.append("FC=" + self.spec["mpi"].mpifc)
+        args.append("F77=" + self.spec["mpi"].mpif77)
```
angus-g, line 48: Chipping away at things, I think these compiler overrides are necessary unless the nci-openmpi package exports its paths differently. Otherwise, we get errors like the following:
```
libtool: link: /g/data/x77/ahg157/spack/lib/spack/env/intel/icpc -g -O2 -g -O2 -fPIC -std=c++11 -D_DO_NOT_LOAD_GLOBALS_ -g -O2 -o .libs/issm_slc.exe main/issm_slc-issm_slc.o  ./.libs/libISSMCore.so -L/g/data/x77/ahg157/spack/opt/spack/linux-rocky8-cascadelake/intel-2021.10.0/petsc-3.19.4-tvawatn/lib -L/g/data/x77/ahg157/spack/opt/spack/linux-rocky8-cascadelake/intel-2021.10.0/mumps-5.5.1-fnmjlzj/lib -L/g/data/x77/ahg157/spack/opt/spack/linux-rocky8-cascadelake/intel-2021.10.0/metis-5.1.0-4gtjfh2/lib -L/apps/openmpi/4.1.4/lib ./.libs/libISSMOverload.so /g/data/x77/ahg157/spack/opt/spack/linux-rocky8-cascadelake/intel-2021.10.0/m1qn3-3.3-vjufbp6/lib/libm1qn3.a /g/data/x77/ahg157/spack/opt/spack/linux-rocky8-cascadelake/intel-2021.10.0/m1qn3-3.3-vjufbp6/lib/libddot.a -lpetsc -ldl -ldmumps -lmumps_common -lpord -lparmetis -lmetis -Wl,-rpath,/g/data/x77/ahg157/spack/opt/spack/linux-rocky8-cascadelake/intel-2021.10.0/issm-develop-ddhchuj/lib
/bin/ld: main/issm-issm.o: undefined reference to symbol 'MPI_Comm_set_attr'
//apps/openmpi/4.1.4/lib/libmpi.so.40: error adding symbols: DSO missing from command line
make[3]: *** [Makefile:5068: issm.exe] Error
```

It's trying to do the MPI linking itself, but using `icpc`, and completely missing all the weird stuff the actual compilers set.

aidanheerdegen: > I think these compiler overrides are necessary unless the `nci-openmpi` package exports its paths differently

Do you think that is something the `nci-openmpi` package should do? Otherwise this fix might need to be done for many packages. 

Is there a standard** for this sort of thing?


** generally assumed way of doing things

angus-g: It's pretty much detailed here. Following that, I should probably set `CC` and go in the environment, rather than just as arguments to configure. https://spack.readthedocs.io/en/latest/packaging_guide.html#mpi-support-in-spack

angus-g: Alright, I did try that recommendation, and no, it doesn't work. I think passing the compilers through to configure is the way to do it. I don't know that there's a "standard", but there's a mix between packages defining the compilers in the configure stage, or in the environment at the install stage. So I think both are legitimate approaches.

harshula: Let's go with what works, for now.

harshula (approved):
TODO: do `from spack.package import install, mkdirp` instead of `from spack.package import *`

aidanheerdegen (approved):
LGTM.

Thanks again @angus-g

PR #44, opened 2023-09-25
Python PR checks

### LINKED ISSUE #38: Use python syntax/semantic/formatting checks on opened PRs for `spack_packages`
@harshula has noted that quick checks on `spack_packages` for python-related syntax/semantic checks would be beneficial during the PR process.

CodeGat:
Adding a workflow for python syntax checks on PR. 

Closes #38 !

### DIFF
--- a/.github/workflows/py-checks.yml
+++ b/.github/workflows/py-checks.yml
@@ -1,0 +1,21 @@
+name: Python Checks
+on:
+  pull_request:
+    paths:
+      - '**.py'
+  push:
+    branches:
+      - main
+    paths:
+      - '**.py'
+jobs:
+  format-syntax-check:
+    name: check formatting and syntax
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: chartboost/ruff-action@v1
+        with:
+          src: "./packages"
+          args: "check --format github --preview --config .ruff.toml"
+
--- a/.github/workflows/py-checks.yml
+++ b/.github/workflows/py-checks.yml
@@ -8,14 +8,18 @@ on:
       - main
     paths:
       - '**.py'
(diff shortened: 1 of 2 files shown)


.github/workflows/py-checks.yml
```diff
@@ -0,0 +1,21 @@
...
+jobs:
+  format-syntax-check:
+    name: check formatting and syntax
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: chartboost/ruff-action@v1
+        with:
+          src: "./packages"
+          args: "check --format github --preview --config .ruff.toml"
```
harshula, line 20: Waiting to see if there's a way to only analyse the files that have changed.

CodeGat: Hi @harshula, just getting back to this now. This is possible, but what was the rationale for only checking files that have changed? Wouldn't it be better to have the entire project to be in compliance with ruff at the start rather than a subset?

harshula: For each PR, we should only be analysing the files that change.

CodeGat: Fair enough, will add the change. Although, errors can be introduced into files even if they are not changed themselves. Consider what happens if you change the signature of a function that is called from a file that is not changed in the PR. This error would not be picked up if we are only checking files that are changed.

harshula: Check if ruff will actually pick up such an error. That is useful.

CodeGat: Looks like ruff doesn't pick up those kinds of cross-file TypeErrors after all :D Added a step that will only run ruff over changed python files

harshula (approved):

PR #51, opened 2023-11-06
mom5: new switch --no_version is required

harshula:
The MOM5 commit d7ba13a3f364ce130b6ad0ba813f01832cada7a2 requires the --no_version switch to avoid git hashes being embedded in the binary.


aidanheerdegen (approved):
LGTM

PR #53, opened 2023-11-13
access-om2: Added link to git repo, removed hardcoded version string

CodeGat:
In this PR, we remove the hardcoded `version("2023.0.001-1")` and add in the git repo, so we can associate a particular build of access-om2 with a git commit (aka, so we can have a `spack install access-om2@git.2023.11.20` for example).

### DIFF
--- a/packages/access-om2/package.py
+++ b/packages/access-om2/package.py
@@ -12,9 +12,9 @@ class AccessOm2(BundlePackage):
 
     homepage = "https://www.access-nri.org.au"
 
-    maintainers = ["harshula"]
+    git = "https://github.com/ACCESS-NRI/ACCESS-OM2.git"
 
-    version("2023.01.001-1")
+    maintainers = ["harshula"]
 
     variant("deterministic", default=False, description="Deterministic build.")
 
--- a/packages/access-om2/package.py
+++ b/packages/access-om2/package.py
@@ -16,6 +16,8 @@ class AccessOm2(BundlePackage):
 
     maintainers = ["harshula"]
 
+    version("latest")
+
     variant("deterministic", default=False, description="Deterministic build.")
 
     depends_on("libaccessom2+deterministic", when="+deterministic")


aidanheerdegen:
This change makes it simpler/easier to update to a new version of the package because it doesn't require a change in this SPD?

I suppose that has significant advantages for our CI/CD testing pipeline as it doesn't require rebuilding and deploying the dependency containers as frequently?

CodeGat:
Yes. Especially for CD as right now every version of ACCESS-OM2 will be `2023.01.001-1` to spack. The other alternative was to not use the ACCESS-OM2 SBD entirely, instead having the spack.yaml being a set of the dependencies already noted in the SBD.

harshula (approved):

harshula (changes requested):
We need to have a single default version(), same as the SPDs that we discussed, otherwise this happens:
```
$ spack install access-om2
==> Error: concretization failed for the following reasons:

   1. Cannot select a single "version" for package "access-om2"
```
Have a look at the SPDs and pick something you think is appropriate.

CodeGat:
Hi @harshula Alternatively, we always give an explicit version on installation I.e. `access-om2@2023.11.22`. 

I'm against the idea of defaults personally. Either the default changes or it remains static. 

If the default is dynamic, users may install the default version every time not knowing it might possibly change, or what the default version is.

If the default is static...is there such thing as a sensible default that isn't just the latest released version? In that case, every deployment would require an update to the SBD. I don't really know what any other sensible defaults would exist that wouldn't require updates each time there was a release.

harshula:
https://github.com/ACCESS-NRI/spack_packages/issues/32#issuecomment-1794014603

harshula:
How about using a string like `latest` or `newest`. We could then use the same string in the SPDs.

harshula (approved):

PR #56, opened 2023-11-29
mom5: print the generated config file to stdout.

harshula:
The cice5 SPD already does this. This code was in the same `development` commit as mom5's `netcdf-c~mpi` dependency testing code and did not reach `main`


CodeGat (approved):
Looks fine to me (except for the formatting errors)

Issue #61, opened 2024-02-05
libaccesom2: Build failure when using Spack v0.21 and nci-openmpi


harshula:
While testing Spack v0.21 on Gadi, the following happened when using `nci-openmpi`:
```
[gadi test-v0.21-nci-openmpi]$ spack install access-om2 ^netcdf-c@4.7.4 ^netcdf-fortran@4.5.2 ^parallelio@2.5.2 ^nci-openmpi@4.0.2 %intel@19.0.5.281

CMake Error in /[...]/tmp/spack-stage/spack-stage-libaccessom2-master-q7h257hjh2fom5c7zgc6annmrzce2pm5/spack-build-q7h257h/CMakeFiles/CMakeTmp/CMakeLists.txt:
  Imported target "MPI::MPI_Fortran" includes non-existent path

    "/apps/openmpi/4.0.2/include/Intel:-I/apps/openmpi/4.0.2/include/Intel"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.


See build log for details:
  /[...]/tmp/spack-stage/spack-stage-libaccessom2-master-q7h257hjh2fom5c7zgc6annmrzce2pm5/spack-build-out.txt

==> Warning: Skipping build of mom5-master-5vptfbmf7kqenqdzmdeuv6udjmitb6bo since libaccessom2-master-q7h257hjh2fom5c7zgc6annmrzce2pm5 failed
==> Warning: Skipping build of access-om2-latest-lzin4rut23ztsvtlq44xc5lre6sddnt4 since mom5-master-5vptfbmf7kqenqdzmdeuv6udjmitb6bo failed
==> Warning: Skipping build of cice5-master-3rqpsj5jag2qvezxs6wlrm5l3ounhwmd since libaccessom2-master-q7h257hjh2fom5c7zgc6annmrzce2pm5 failed
==> Error: access-om2-latest-lzin4rut23ztsvtlq44xc5lre6sddnt4: Package was not installed
==> Error: Installation request failed.  Refer to reported errors for failing package(s).
```

Testing Spack v0.21 in a Docker image with `openmpi`:
```
[root@d038f8bf660a /]# spack --version
0.21.1 (e30fedab102f9281a220fb4fae82e3f8c43a82ac)

[...]

PR #66, opened 2024-02-20
libaccessom2: explicitly set CMake compiler variables

harshula:
Details can be found in https://github.com/ACCESS-NRI/spack-packages/issues/61 .


aidanheerdegen (approved):
LGTM.

Worth stating explicitly that this change forces CMake to use the MPI compiler wrappers which add localised flags to find libraries and includes.

Issue #68, opened 2024-02-27
Spack package for GCOM 4.5


penguian:
See Issue #59. Create the Spack package `gcom4`, based on `gcom`, but using the https://github.com/ACCESS-NRI/GCOM repository instead of `MOSRS/gcom`. This is for a dependency of the `um7` Spack package to be used for ESM 1.5.

PR #69, opened 2024-02-27
Add gcom4 package

penguian:
Closes Issue #68 . See that issue for a description of the change requested.

This package applies patches to files in https://github.com/ACCESS-NRI/GCOM/tree/dev/Share/gcom4.5_access_config
as per https://github.com/coecms/access-esm-build-gadi/blob/master/Makefile rather than relying on updating the upstream repository.

The package has been tested to the point where the install completes.

### DIFF
--- a/packages/gcom4/package.py
+++ b/packages/gcom4/package.py
@@ -21,7 +21,7 @@ class Gcom4(Package):
     homepage = "https://code.metoffice.gov.uk/trac/gcom"
     git = "git@github.com:ACCESS-NRI/GCOM.git"
 
-    maintainers = ["penguian"]
+    maintainers("penguian")
 
     version("4.5", branch="dev")
     build_directory = join_path("Share", "gcom4.5_access_config")
--- a/packages/gcom4/package.py
+++ b/packages/gcom4/package.py
@@ -31,6 +31,23 @@ class Gcom4(Package):
     depends_on("mpi", when="+mpi")
 
 
+    def gcom_machine(self, spec):
+        """
+        Determine the machine configuration name
+        """
+        if spec.satisfies("%intel"):
+            mach_c = "ifort"
+        elif spec.satisfies("%gcc"):
+            mach_c = "gfortran"
+        else:
+            raise NotImplementedError("Unknown compiler")
+        if "+mpi" in spec:
+            mach_m = "openmpi"
+        else:
+            mach_m = "serial"
+        return f"nci_{mach_c}_{mach_m}"
+
+
     def patch(self):
         """
         Perform the equivalent of the following sed commands:
@@ -42,10 +59,11 @@ class Gcom4(Package):
             filter_file(
                 r"build\.target\{ns\}.*", "#",
                 join_path("fcm-make", "gcom.cfg"))
-            for mach_m in ["openmpi", "serial"]:
+            if self.spec.satisfies("%intel"):
+                machine = self.gcom_machine(self.spec)
                 filter_file(
                     r"-openmp", "-qopenmp",
-                        join_path("fcm-make", "machines", f"nci_ifort_{mach_m}.cfg"))
+                        join_path("fcm-make", "machines", f"{machine}.cfg"))
 
             
     def build(self, spec, prefix):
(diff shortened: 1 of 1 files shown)


penguian:
I could add the following to package.py but have chosen not to at the moment:

1. Ensure that the dependent `fcm` executable is at the head of `$PATH` :
```         
    def setup_build_environment(self, env):
        env.prepend_path("PATH", self.spec["fcm"].prefix.bin)
```

2. Alternatively, use the `fcm` executable defined by `join_path(spec["fcm"].prefix.bin, "fcm"))`:
```

        def make_executable(spec, name):
            return Executable(join_path(spec[name].prefix.bin, name))

        fcm = make_executable(spec, "fcm")
        ...
```
But I would rather submit the package as-is, and debate the merits of such changes later.

Also, I would like to set the default compiler to `"%intel"` but I understand that this is not necessary.

packages/gcom4/package.py
```diff
@@ -0,0 +1,88 @@
...
+class Gcom4(Package):
+    """
+    GCOM is a wrapper around multiprocessing libraries such as MPI
+    Gcom4 is a package for older versions of GCOM as used by UM vn7.3 in ACCESS ESM 1.5
+    """
+
+    homepage = "https://code.metoffice.gov.uk/trac/gcom"
+    git = "git@github.com:ACCESS-NRI/GCOM.git"
+
+    maintainers = ["penguian"]
```
harshula, line 24: Spack is changing this to a function. See https://spack.readthedocs.io/en/latest/packaging_guide.html#maintainers

packages/gcom4/package.py
```diff
@@ -0,0 +1,88 @@
...
+        with fs.working_dir(self.build_directory):
+            filter_file(
+                r"build\.target\{ns\}.*", "#",
+                join_path("fcm-make", "gcom.cfg"))
+            filter_file(
+                r"-openmp", "-qopenmp",
+                join_path("fcm-make", "machines", "nci_ifort_openmpi.cfg"))
+
+            
+    def install(self, spec, prefix):
```
harshula, line 50: Can you have separate build() function?

packages/gcom4/package.py
penguian: Will do.

packages/gcom4/package.py
penguian: I will test this.

penguian: It does not work. Also, [the doc](https://spack.readthedocs.io/en/latest/packaging_guide.html) says 

> `PackageBase` only supports the `install` phase while the `AutotoolsPackage` and `MakefilePackage` support both install and build phases.

I worked around this by calling `self.build()` from `install`. Is that what you want?
```
$ spack install gcom4%intel
[+] /g/data/tm70/pcl851/src/ACCESS-NRI/release/linux-rocky8-x86_64/intel-19.0.5.281/fcm-2021.05.0-i4kamrjix5ggowbmvz26e6a7shh7xxqv
==> openmpi@4.0.2 : has external module in ['openmpi/4.0.2']
[+] /apps/openmpi/4.0.2 (external openmpi-4.0.2-4jwtg3nt7p3zsod4rvn3i3jyyop3arpe)
==> Installing gcom4-4.5-tshdzy43m2k7g6zf3gcysfgrnxcwpkqj
==> No binary for gcom4-4.5-tshdzy43m2k7g6zf3gcysfgrnxcwpkqj found: installing from source
==> Ran patch() for gcom4
==> gcom4: Executing phase: 'install'
==> Warning: Module file /g/data/tm70/pcl851/src/ACCESS-NRI/release/modules/linux-rocky8-x86_64/gcom4/4.5 exists and will not be overwritten
==> gcom4: Successfully installed gcom4-4.5-tshdzy43m2k7g6zf3gcysfgrnxcwpkqj
  Stage: 10.49s.  Install: 1m 37.85s.  Post-install: 0.22s.  Total: 1m 51.09s
  ```

harshula: I was only wondering if build() was supported. However, what you did makes it clearer.

packages/gcom4/package.py
penguian: Done.

harshula:
Hi @penguian , Are there more changes planned for the `patch()` function?

penguian:
@harshula I have made my changes and the PR is again ready for review.

harshula (approved):

PR #79, opened 2024-03-21
Add restart_repro variant to access-om2 (default true)

aidanheerdegen:
Add +restart_repro variant to access-om2 (default true)

As the variant only needs to be propagated to `mom5` it doesn't add a heap of ugly combinatorial boilerplate.

Closes #78

PR #82, opened 2024-03-21
Revert "Add restart_repro variant to access-om2 (default true)"

aidanheerdegen:
Reverts ACCESS-NRI/spack-packages#79


harshula:
We decided to do https://github.com/ACCESS-NRI/spack-packages/pull/83 instead.

PR #84, opened 2024-03-26
Initial package for dummygrib

penguian:
The repository at https://github.com/coecms/dummygrib includes a `Makfile` without an `install` target. The package for `dummygrib` is based on `MakefilePackage` with an added `install` member function. Tested the package, and the library builds and installs with both `gcc-8.5.0` and `intel-19.0.5.281` compilers.

### DIFF
--- a/packages/dummygrib/package.py
+++ b/packages/dummygrib/package.py
@@ -1,0 +1,32 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# Copyright 2024 ACCESS-NRI
+# Based on https://github.com/coecms/access-esm-build-gadi/blob/master/Makefile
+#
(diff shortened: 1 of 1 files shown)


packages/dummygrib/package.py
```diff
@@ -0,0 +1,32 @@
...
+from spack.package import *
+
+
+class Dummygrib(MakefilePackage):
+    """
+    dummygrib is a Dummy GRIB library to use with the Met Office Unified Model.
+    """
+
+    homepage = "https://github.com/coecms/dummygrib"
+    git = "git@github.com:coecms/dummygrib.git"
```
harshula, line 18: If this is not a private repository, please use `https://github.com/coecms/dummygrib.git`.

penguian: Done.

harshula (approved):

penguian:
Merging.

PR #90, opened 2024-04-12
Added `access-om3-virtual` bundle definition

### LINKED ISSUE #89: Create `access-om3` SBD that houses the equivalent SPD
See https://github.com/ACCESS-NRI/ACCESS-OM3/issues/3#issuecomment-2044222491

Essentially, we will need an overarching `access-om3` SBD. This will be used to version the entirety of an `access-om3` deployment release, including not only the `access-om3` package, but the dependencies, the versions of `spack-{packages,config}` and the `spack.yaml` itself. This is similar to https://github.com/ACCESS-NRI/access-om2. 

Currently we are leaning towards having the naming scheme as follows:
* SBD: `access-om3-virtual`
* SPD: `access-om3`

References ACCESS-NRI/ACCESS-OM3#3
References ACCESS-NRI/ACCESS-OM3#2

CodeGat:
Closes #89 
References https://github.com/ACCESS-NRI/ACCESS-OM3/issues/3

### DIFF
--- a/packages/access-om3-virtual/package.py
+++ b/packages/access-om3-virtual/package.py
@@ -1,0 +1,28 @@
+# Copyright 2013-2023 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
+#
+# Copyright 2024 ACCESS-NRI
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import BundlePackage, version, depends_on
+
+
+class AccessOm3Virtual(BundlePackage):
+    """
+    ACCESS-OM3 bundle containing the ACCESS-OM3 Package.
+
+    This is used to version the entirety of a released deployment, including
+    the package, it's dependencies, and the version of
+    spack-packages/spack-config that it is bundled with
+    """
+
+    homepage = "https://www.access-nri.org.au"
+
+    git = "https://github.com/ACCESS-NRI/ACCESS-OM3.git"
+
+    maintainers = ["harshula"]
+
+    version("latest")
+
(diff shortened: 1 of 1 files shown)


packages/access-om3-virtual/package.py
```diff
@@ -0,0 +1,29 @@
...
+
+    homepage = "https://www.access-nri.org.au"
+
+    git = "https://github.com/ACCESS-NRI/ACCESS-OM3.git"
+
+    maintainers = ["harshula"]
+
+    version("latest")
+
+    # FIXME: How do we inject variants into the package from the SBD?
```
harshula, line 28: Example from `access-om2`:
```
    variant("deterministic", default=False, description="Deterministic build.")

    depends_on("libaccessom2+deterministic", when="+deterministic")
    depends_on("libaccessom2~deterministic", when="~deterministic")
    depends_on("cice5+deterministic", when="+deterministic")
    depends_on("cice5~deterministic", when="~deterministic")
    depends_on("mom5+deterministic", when="+deterministic")
    depends_on("mom5~deterministic", when="~deterministic")
```

CodeGat: Although I'm wondering if we need to have those variants in there, since they can be specified in the `spack.yaml` like so:
```yaml
spack:
  specs:
    - access-om3-virtual@git.2024.04.0
  packages:
    access-om3:
      require:
        - '@git.v0.2.0'
        - '<variants here>
    # more packages...
```

harshula: For non-environment installs the SBD would need the variants that you insert in the spack.yaml. It's best to keep them synchronised until we have a better option.

harshula (approved):

Issue #91, opened 2024-04-17
Copy required packages from `COSIMA/spack_packages`, rename root `access-om3` SBD


CodeGat:
The current `0.x.0` `spack.yaml` from `COSIMA` requires spack packages that we do not have in our repository. 
Currently, the ones that will have to be copied are:
* `fms`
* `fortranxml`

Furthermore, in the `spack.yaml` we are working with in https://github.com/ACCESS-NRI/ACCESS-OM3/pull/5, we initially had our root SBD element as `access-om3-virtual`, with the `access-om3` SPD being it's only dependency. 
We are looking to rename the following packages, in order to keep the names that the user will interact with more simple:
* `access-om3` (SPD) renamed to `access-om3-cmake` (open to change the name)
* `access-om3-virtual` (SBD) renamed to `access-om3`

PR #93, opened 2024-04-17
Copy packages required for `OM3` from `COSIMA/spack_packages`

CodeGat:
In this PR:
* Added the `COSIMA/spack_packages` packages required for the OM3 `spack.yaml` - ~`fms`~, `fortranxml` at https://github.com/COSIMA/spack_packages/commit/a65eaf418730f7f132be3061e64b5f8c05c475ca

References #91


micaeljtoliveira:
Probably it would be good to submit the changes to the FMS package upstream.

packages/fms/package.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+class Fms(CMakePackage):
+    """GFDL's Flexible Modeling System (FMS) is a software environment
+    that supports the efficient development, construction, execution,
+    and scientific interpretation of atmospheric, oceanic, and climate
+    system models."""
+
+    homepage = "https://github.com/NOAA-GFDL/FMS"
+    url = "https://github.com/NOAA-GFDL/FMS/archive/refs/tags/2022.04.tar.gz"
```
harshula, line 16: Hopefully we can implement something like `url_for_version()` in the future instead of hardcording this. e.g. This is what we do in the MOM5 SPD:
```
        return "https://github.com/ACCESS-NRI/mom5/tarball/{0}".format(version)
```

harshula:
There's an upstream version of fms at https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/fms/package.py

harshula:
Hi @micaeljtoliveira , Do you use `large_file` and/or `internal_file_nml` variants?

harshula:
Hi @CodeGat , Probably drop the copying of `fms` SPD. Spack v0.21 `fms` is only missing two variants and shouldn't impact your CI.

micaeljtoliveira:
> Hi @micaeljtoliveira , Do you use large_file and/or internal_file_nml variants?

Yes. Otherwise I wouldn't have bothered adding them ;)

harshula:
Hi @micaeljtoliveira , Can you please submit the patch upstream? They are default false in the SPD. Can you please elaborate how you use them? Are they default true in FMS' build system?

packages/fortranxml/package.py
```diff
@@ -0,0 +1,44 @@
...
+
+class Fortranxml(AutotoolsPackage):
+    """Fortran XML (FoX) s a library, written entirely in Fortran, designed to allow
+    the easy use of XML from Fortran programs. FoX caters for both XML output
+    and input.
+    """
+
+    homepage = "https://github.com/andreww/fox"
+    git = "https://github.com/andreww/fox.git"
+    url = "https://github.com/andreww/fox/archive/refs/tags/4.1.2.tar.gz"
```
harshula, line 17: Might be better to implement `url_for_version()`.

CodeGat: Done

micaeljtoliveira:
> Hi @micaeljtoliveira , Can you please submit the patch upstream? They are default false in the SPD. Can you please elaborate how you use them? Are they default true in FMS' build system?

Yes, sure, I can submit upstream. `+large_file` is set explicitly in the environment definition. Looks like `internal_file_nml` is not set, but I'm afraid I don't remember exactly why. The only thing I'm quite sure is that both are set by in CESM, which was used as a template for OM3.

packages/fortranxml/package.py
```diff
@@ -0,0 +1,46 @@
...
+class Fortranxml(AutotoolsPackage):
+    """Fortran XML (FoX) s a library, written entirely in Fortran, designed to allow
+    the easy use of XML from Fortran programs. FoX caters for both XML output
+    and input.
+    """
+
+    homepage = "https://github.com/andreww/fox"
+    git = "https://github.com/andreww/fox.git"
+
+    maintainers = ["andreww"]
```
harshula, line 18: https://spack.readthedocs.io/en/latest/packaging_guide.html#maintainers

packages/fortranxml/package.py
```diff
@@ -0,0 +1,46 @@
...
+
+    version("4.1.2", sha256="dea0adc9cc035238fa9cdba42f2bf56481e3a64ac8aa0aece9119f127f71d4e7")
+
+    flag_handler = AutotoolsPackage.build_system_flags
+
+    def url_for_version(self, version):
+        return "https://github.com/andreww/fox/tarball/{0}".format(version)
+
+    def install(self, spec, prefix):
+        super(AutotoolsPackage, self).install(spec, prefix)
```
harshula, line 28: Hi @micaeljtoliveira , I haven't seen this before in an SPD. Should it be `super().install(spec, prefix)`? This might be something we should do in our existing SPDs, so let's have a chat about it next week.

packages/fortranxml/package.py
CodeGat: I take it you will be the maintainer, then, @harshula ? :laughing:

harshula: Sure.

packages/fms/package.py
CodeGat: No longer relevant because there is no `fms` package in the repo.

packages/fortranxml/package.py
CodeGat: Done

harshula (approved):
Approving for merging. We'll fix any remaining issues later.

PR #94, opened 2024-04-17
Renamed `access-om3` -> `access-om3-nuopc`, `access-om3-virtual` -> `access-om3`

CodeGat:
In this PR:
* Renamed SBD from `access-om3-virtual` to `access-om3`
* Renamed SPD from `access-om3` to `access-om3-nuopc` (Open to changing this name. This also may affect some COSIMA stuff, so I'd like to get @micaeljtoliveira to review and approve)

Tagging @aidanheerdegen as well.

References #91


micaeljtoliveira:
I'm still trying to wrap my head around the need to have a bundle package for something that is definitely not a bundle. Really feels like you are trying to hammer a square peg into round hole. Anyway, this only requires some minor changes in the COSIMA spack instance, so it's fine from that point of view.

micaeljtoliveira (approved):

CodeGat:
Thanks for the feedback, Micael. Ultimately it's because we need to have something in the `spack.yaml` to version the entirety of the deployment, as we'd made a decision regarding the CI that we'd rather have it interrogate/use a version in the `spack.yaml` rather than something like a `VERSION` file. We can't use the `access-om3-nuopc` package because the version string there points to the OM3 codebase (`0.2.0`, etc). I do see your point regarding the bundle-package-that-isn't-really-a-bundle, though.

PR #106, opened 2024-05-13
Add an ACCESS-ESM1.5 version

### LINKED ISSUE #104: mom5: Add an access-esm1.5 version
Add an `access-esm1.5` version of MOM5, using the https://github.com/ACCESS-NRI/MOM5/tree/access-esm1.5 branch, possibly invoked by giving a new value of `ACCESS-ESM1.5` for the `type` option. The build of the `access-esm1.5` variant should match https://github.com/penguian/access-esm-build-gadi/blob/master/Makefile, which actually uses the `ACCESS-CM` `type` option.
```
bin/mom5xx : src/mom5 $(ENVFILE) lib/oasis | bin
	source $(ENVFILE); cd $</exp; ./MOM_compile.csh --platform=access-cm2 --type=ACCESS-CM
	cp src/mom5/exec/access-cm2/ACCESS-CM/fms_ACCESS-CM.x $@
```

penguian:
Closes #104 
The build of
```
spack install mom5@access-esm1.5%intel@19.0.5.281 arch=linux-rocky8-x86_64
```
completes successfully, and running https://github.com/ACCESS-NRI/access-esm1.5-configs/tree/pre-industrial with appropriate changes to executable locations
```
$ grep "    exe:" config.yaml
      exe: /g/data/tm70/pcl851/src/ACCESS-NRI/release/linux-rocky8-x86_64/intel-19.0.5.281/um7-7.3-tn5nef76eygrvbstm3ionmid6hysrij2/bin/um_hg3.exe
      exe: /g/data/tm70/pcl851/src/ACCESS-NRI/release/linux-rocky8-x86_64/intel-19.0.5.281/mom5-access-esm1.5-qgaq7scgvrlvm77yvqrhllzbl7657pzl/bin/fms_ACCESS-CM.x
      exe: /g/data/tm70/pcl851/src/ACCESS-NRI/release/linux-rocky8-x86_64/intel-19.0.5.281/cice4-access-esm1.5-ktkbqdnat4etwrh6v3zavmricogmin5e/bin/cice_access_360x300_12x1_12p.exe
```
also completes, but does not produce bitwise identical MOM5 output.

### DIFF
--- a/packages/mom5/package.py
+++ b/packages/mom5/package.py
@@ -14,12 +14,13 @@ class Mom5(MakefilePackage):
     homepage = "https://www.access-nri.org.au"
     git = "https://github.com/ACCESS-NRI/mom5.git"
 
-    maintainers = ["harshula", "penguian"]
+    maintainers("harshula", "penguian")
 
-    version("master", branch="master")
     version("access-esm1.5", branch="access-esm1.5")
+    version("master", branch="master")
 
     variant("deterministic", default=False, when="@master", description="Deterministic build.")
+    variant("esm-type", default="ACCESS-CM", when="@access-esm1.5", description="Build MOM5 to support a particular ESM use case.", values=("ACCESS-CM", ), multi=False)
     variant("type", default="ACCESS-OM", when="@master", description="Build MOM5 to support a particular use case.", values=("ACCESS-CM", "ACCESS-ESM", "ACCESS-OM", "ACCESS-OM-BGC", "MOM_solo"), multi=False)
     variant("restart_repro", default=True, description="Reproducible restart build.")
     variant("optimisation_report", default=False, when="@master", description="Generate optimisation reports.")
@@ -29,11 +30,11 @@ class Mom5(MakefilePackage):
     depends_on("netcdf-fortran@4.5.2:")
     depends_on("netcdf-c@4.7.4:")
     depends_on("datetime-fortran", when="@master")
-    depends_on("oasis3-mct+deterministic", when="+deterministic@master")
-    depends_on("oasis3-mct~deterministic", when="~deterministic@master")
     depends_on("oasis3-mct@access-esm1.5", when="@access-esm1.5")
-    depends_on("libaccessom2+deterministic", when="+deterministic@master")
-    depends_on("libaccessom2~deterministic", when="~deterministic@master")
+    depends_on("oasis3-mct+deterministic", when="+deterministic")
+    depends_on("oasis3-mct~deterministic", when="~deterministic")
+    depends_on("libaccessom2+deterministic", when="+deterministic")
+    depends_on("libaccessom2~deterministic", when="~deterministic")
 
     phases = ["edit", "build", "install"]
 
@@ -93,16 +94,16 @@ FFLAGS += -DGFORTRAN
 
 #
 FFLAGS_OPT = -O2
-FFLAGS_REPRO =
-FFLAGS_DEBUG = -O0 -g -W -fbounds-check
-FFLAGS_OPENMP = -fopenmp
+FFLAGS_REPRO = 
+FFLAGS_DEBUG = -O0 -g -W -fbounds-check 
+FFLAGS_OPENMP = -fopenmp 
 FFLAGS_VERBOSE =
 
 CFLAGS := -D__IFC {incs}
 CFLAGS += $(shell nc-config --cflags)
 CFLAGS_OPT = -O2
 CFLAGS_OPENMP = -fopenmp
-CFLAGS_DEBUG = -O0 -g
+CFLAGS_DEBUG = -O0 -g 
 
 # Optional Testing compile flags.  Mutually exclusive from DEBUG, REPRO, and OPT
 # *_TEST will match the production if no new option(s) is(are) to be tested.
@@ -111,7 +112,7 @@ CFLAGS_TEST = -O2
 
 LDFLAGS :=
 LDFLAGS_OPENMP := -fopenmp
-LDFLAGS_VERBOSE :=
+LDFLAGS_VERBOSE := 
 
 ifneq ($(REPRO),)
 CFLAGS += $(CFLAGS_REPRO)
@@ -475,7 +476,7 @@ TMPFILES = .*.m *.T *.TT *.hpm *.i *.lst *.proc *.s
                     build.add_default_env("REPRO", "true")
                 build(
                     "--type",
-                    "ACCESS-CM",
+                    self.spec.variants["esm-type"].value,
                     "--platform",
                     self._platform,
                     "--no_environ"
@@ -501,10 +502,8 @@ TMPFILES = .*.m *.T *.TT *.hpm *.i *.lst *.proc *.s
     def install(self, spec, prefix):
 
         mkdirp(prefix.bin)
-        if "@access-esm1.5" in self.spec:
-            type_value = "ACCESS-CM"
-        else:
-            type_value = self.spec.variants["type"].value
+        name = "esm-type" if "@access-esm1.5" in self.spec else "type"
+        type_value = self.spec.variants[name].value
 
         install(
             join_path(
--- a/packages/mom5/package.py
+++ b/packages/mom5/package.py
@@ -21,9 +21,9 @@ class Mom5(MakefilePackage):
 
     variant("type", default="ACCESS-OM", description="Build MOM5 to support a particular use case.", values=("ACCESS-CM", "ACCESS-ESM", "ACCESS-OM", "ACCESS-OM-BGC", "MOM_solo"), multi=False)
     variant("restart_repro", default=True, description="Reproducible restart build.")
-    with when("@master"):
-        variant("deterministic", default=False, description="Deterministic build.")
-        variant("optimisation_report", default=False, description="Generate optimisation reports.")
+    # The following two variants are not applicable when version is "access-esm1.5":
+    variant("deterministic", default=False, description="Deterministic build.")
+    variant("optimisation_report", default=False, description="Generate optimisation reports.")
     with when("@access-esm1.5"):
         variant("type", default="ACCESS-CM", description="Build MOM5 to support a particular ESM use case.", values=("ACCESS-CM", ), multi=False)
 
@@ -31,7 +31,8 @@ class Mom5(MakefilePackage):
     depends_on("mpi")
     depends_on("netcdf-fortran@4.5.2:")
     depends_on("netcdf-c@4.7.4:")
-    with when("@master"):
+    # Spack does not have a spec syntax for NOT "@access-esm1.5", so use version ranges instead
+    with when("@:access-esm0,access-esm2:"):
         depends_on("datetime-fortran")
         depends_on("oasis3-mct+deterministic", when="+deterministic")
         depends_on("oasis3-mct~deterministic", when="~deterministic")
@@ -90,7 +91,7 @@ OPENMP =
 
 MAKEFLAGS += --jobs=$(shell grep '^processor' /proc/cpuinfo | wc -l)
 
-FPPFLAGS :=
+FPPFLAGS := 
 
 FFLAGS := -fcray-pointer -fdefault-real-8 -ffree-line-length-none -fno-range-check -Waliasing -Wampersand -Warray-bounds -Wcharacter-truncation -Wconversion -Wline-truncation -Wintrinsics-std -Wsurprising -Wno-tabs -Wunderflow -Wunused-parameter -Wintrinsic-shadow -Wno-align-commons -fallow-argument-mismatch -fallow-invalid-boz
 FFLAGS += {incs}
@@ -100,8 +101,8 @@ FFLAGS += -DGFORTRAN
 FFLAGS_OPT = -O2
 FFLAGS_REPRO = 
 FFLAGS_DEBUG = -O0 -g -W -fbounds-check 
-FFLAGS_OPENMP = -fopenmp 
-FFLAGS_VERBOSE =
+FFLAGS_OPENMP = -fopenmp
+FFLAGS_VERBOSE = 
 
 CFLAGS := -D__IFC {incs}
 CFLAGS += $(shell nc-config --cflags)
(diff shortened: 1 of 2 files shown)


packages/mom5/package.py
```diff
@@ -14,24 +14,26 @@ class Mom5(MakefilePackage):
...
-    maintainers = ["harshula"]
+    maintainers = ["harshula", "penguian"]
 
     version("master", branch="master")
+    version("access-esm1.5", branch="access-esm1.5")
 
-    variant("deterministic", default=False, description="Deterministic build.")
-    variant("type", default="ACCESS-OM", description="Build MOM5 to support a particular use case.", values=("ACCESS-CM", "ACCESS-ESM", "ACCESS-OM", "ACCESS-OM-BGC", "MOM_solo"), multi=False)
+    variant("deterministic", default=False, when="@master", description="Deterministic build.")
+    variant("type", default="ACCESS-OM", when="@master", description="Build MOM5 to support a particular use case.", values=("ACCESS-CM", "ACCESS-ESM", "ACCESS-OM", "ACCESS-OM-BGC", "MOM_solo"), multi=False)
```
harshula, line 23: It would be better to make these conditional on when NOT `access-esm1.5`.

packages/mom5/package.py
```diff
@@ -14,24 +14,26 @@ class Mom5(MakefilePackage):
...
     variant("restart_repro", default=True, description="Reproducible restart build.")
-    variant("optimisation_report", default=False, description="Generate optimisation reports.")
+    variant("optimisation_report", default=False, when="@master", description="Generate optimisation reports.")
```
harshula, line 25: It would be better to make these conditional on when NOT `access-esm1.5`.

packages/mom5/package.py
```diff
@@ -14,24 +14,26 @@ class Mom5(MakefilePackage):
...
     depends_on("netcdf-fortran@4.5.2:")
     depends_on("netcdf-c@4.7.4:")
-    depends_on("datetime-fortran")
-    depends_on("oasis3-mct+deterministic", when="+deterministic")
-    depends_on("oasis3-mct~deterministic", when="~deterministic")
-    depends_on("libaccessom2+deterministic", when="+deterministic")
-    depends_on("libaccessom2~deterministic", when="~deterministic")
+    depends_on("datetime-fortran", when="@master")
+    depends_on("oasis3-mct+deterministic", when="+deterministic@master")
+    depends_on("oasis3-mct~deterministic", when="~deterministic@master")
```
harshula, line 33: It would be better to make these conditional on when NOT `access-esm1.5`.

packages/mom5/package.py
```diff
@@ -14,24 +14,26 @@ class Mom5(MakefilePackage):
...
+    depends_on("oasis3-mct@access-esm1.5", when="@access-esm1.5")
+    depends_on("libaccessom2+deterministic", when="+deterministic@master")
+    depends_on("libaccessom2~deterministic", when="~deterministic@master")
```
harshula, line 36: It would be better to make these conditional on when NOT `access-esm1.5`.

packages/mom5/package.py
```diff
@@ -312,33 +470,48 @@ def build(self, spec, prefix):
...
+                    self._platform,
+                    "--no_environ",
+                    "--no_version"
+                )
 
     def install(self, spec, prefix):
 
         mkdirp(prefix.bin)
+        if "@access-esm1.5" in self.spec:
+            type_value = "ACCESS-CM"
```
harshula, line 505: Can you set `self.spec.variants["type"].value = "ACCESS-CM"` much earlier if the version is access-esm1.5.

packages/mom5/package.py
```diff
@@ -14,24 +14,26 @@ class Mom5(MakefilePackage):
     homepage = "https://www.access-nri.org.au"
     git = "https://github.com/ACCESS-NRI/mom5.git"
 
-    maintainers = ["harshula"]
+    maintainers = ["harshula", "penguian"]
```
harshula, line 17: Use `maintainers()`

packages/mom5/package.py
```diff
@@ -14,20 +14,23 @@ class Mom5(MakefilePackage):
     homepage = "https://www.access-nri.org.au"
     git = "https://github.com/ACCESS-NRI/mom5.git"
 
-    maintainers = ["harshula"]
+    maintainers("harshula", "penguian")
 
+    version("access-esm1.5", branch="access-esm1.5")
```
harshula, line 19: The first `version()` is the default, so that has to be `master`.

packages/mom5/package.py
penguian: I have looked at the Spack source code, documentation and issues, and haven't found a way to create the equivalent of
```
variant(name, blah, when="NOT @access-esm1.5", ...)
```
Specifically, there is no NOT for versions. See, for example https://github.com/spack/spack/issues/1975

packages/mom5/package.py
penguian: As above.

packages/mom5/package.py
penguian: As above.

packages/mom5/package.py
penguian: I left off the `@master` since `deterministic` depends on `when="@master"`.

packages/mom5/package.py
penguian: Done, except that I had to use "esm-type". Spack takes the last `variant` call that uses the same name, only keeping `when` from previous `variant` calls. See https://github.com/spack/spack/blob/develop/lib/spack/docs/packaging_guide.rst#overriding-variants

packages/mom5/package.py
penguian: Will do.

penguian: Done.

packages/mom5/package.py
harshula: What happens when we do `mom5@git.<tag>`?

packages/mom5/package.py
harshula: What happens when we do mom5@git.<tag>?

harshula: What happens when we do `mom5@git.<tag>`?

packages/mom5/package.py
penguian: I tested the default with bothe orders, and `access-esm1.5` is always the default. This is possibly because of the `".5"` at the end of the version name.

packages/mom5/package.py
```diff
@@ -14,20 +14,23 @@ class Mom5(MakefilePackage):
...
+    maintainers("harshula", "penguian")
 
     version("master", branch="master")
+    version("access-esm1.5", branch="access-esm1.5")
 
-    variant("deterministic", default=False, description="Deterministic build.")
-    variant("type", default="ACCESS-OM", description="Build MOM5 to support a particular use case.", values=("ACCESS-CM", "ACCESS-ESM", "ACCESS-OM", "ACCESS-OM-BGC", "MOM_solo"), multi=False)
+    variant("deterministic", default=False, when="@master", description="Deterministic build.")
+    variant("esm-type", default="ACCESS-CM", when="@access-esm1.5", description="Build MOM5 to support a particular ESM use case.", values=("ACCESS-CM", ), multi=False)
+    variant("type", default="ACCESS-OM", when="@master", description="Build MOM5 to support a particular use case.", values=("ACCESS-CM", "ACCESS-ESM", "ACCESS-OM", "ACCESS-OM-BGC", "MOM_solo"), multi=False)
```
harshula, line 24:
```suggestion
variant("deterministic", default=False, description="Deterministic build.")
variant("type", default="ACCESS-OM", description="Build MOM5 to support a particular use case.", values=("ACCESS-CM", "ACCESS-ESM", "ACCESS-OM", "ACCESS-OM-BGC", "MOM_solo"), multi=False)
```

packages/mom5/package.py
```diff
@@ -14,20 +14,23 @@ class Mom5(MakefilePackage):
...
     variant("restart_repro", default=True, description="Reproducible restart build.")
-    variant("optimisation_report", default=False, description="Generate optimisation reports.")
+    variant("optimisation_report", default=False, when="@master", description="Generate optimisation reports.")
```
harshula, line 26:
```suggestion
    variant("deterministic", default=False, description="Deterministic build.")
    variant("type", default="ACCESS-OM", description="Build MOM5 to support a particular use case.", values=("ACCESS-CM", "ACCESS-ESM", "ACCESS-OM", "ACCESS-OM-BGC", "MOM_solo"), multi=False)
    variant("optimisation_report", default=False, description="Generate optimisation reports.")

    # https://github.com/spack/spack/blob/develop/lib/spack/docs/packaging_guide.rst#overriding-variants
    with when("@access-esm1.5"):
        variant("type", default="ACCESS-CM", description="Build MOM5 to support a particular ESM use case.", values=("ACCESS-CM", ), multi=False)
```

packages/mom5/package.py
```diff
@@ -312,33 +471,46 @@ def build(self, spec, prefix):
...
+                    self._platform,
+                    "--no_environ",
+                    "--no_version"
+                )
 
     def install(self, spec, prefix):
 
         mkdirp(prefix.bin)
+        name = "esm-type" if "@access-esm1.5" in self.spec else "type"
+        type_value = self.spec.variants[name].value
```
harshula, line 506:
```suggestion
```

packages/mom5/package.py
```diff
@@ -312,33 +471,46 @@ def build(self, spec, prefix):
...
+
         install(
             join_path(
                 "exec",
                 self._platform,
-                self.spec.variants["type"].value,
-                "fms_" + self.spec.variants["type"].value + ".x"
+                type_value,
+                "fms_" + type_value + ".x"
```
harshula, line 513:
```suggestion
                self.spec.variants["type"].value,
                "fms_" + self.spec.variants["type"].value + ".x"
```

packages/mom5/package.py
```diff
@@ -312,33 +471,46 @@ def build(self, spec, prefix):
...
-                self._platform,
-                "--no_environ",
-                "--no_version"
-            )
+            if "@access-esm1.5" in self.spec:
+                if "+restart_repro" in self.spec:
+                    build.add_default_env("REPRO", "true")
+                build(
+                    "--type",
+                    self.spec.variants["esm-type"].value,
```
harshula, line 479:
```suggestion
                    self.spec.variants["type"].value,
```

packages/mom5/package.py
penguian: The Spack documentation describes version comparison here: 
https://spack.readthedocs.io/en/latest/packaging_guide.html#version-comparison
"Infinity versions are not automatically used when determining the latest version of a package unless explicitly required by another package or user." So, for example, if `master` is not explicitly specified, it won't be automatically used.

The Spack documentation describes version selection here: 
https://spack.readthedocs.io/en/latest/packaging_guide.html#version-selection
The largest version in the sort order is used. It is possible to mark a version as `preferred=True`.

packages/mom5/package.py
```diff
@@ -312,33 +471,46 @@ def build(self, spec, prefix):
...
+                # requires the --no_version switch to avoid git hashes being
+                # embedded in the binary.
+                build(
+                    "--type",
+                    self.spec.variants["type"].value,
+                    "--platform",
+                    self._platform,
+                    "--no_environ",
+                    "--no_version"
+                )
```
harshula, line 500:
```suggestion
            if not "@access-esm1.5" in self.spec:
                build.add_default_arg("--no_version")

            if "+optimisation_report" in self.spec:
                build.add_default_env("REPORT", "true")
            if "+restart_repro" in self.spec:
                build.add_default_env("REPRO", "true")

            # The MOM5 commit d7ba13a3f364ce130b6ad0ba813f01832cada7a2
            # requires the --no_version switch to avoid git hashes being
            # embedded in the binary.
            build(
                "--type",
                self.spec.variants["type"].value,
                "--platform",
                self._platform,
                "--no_environ",
            )
```

packages/mom5/package.py
```diff
@@ -14,24 +14,31 @@ class Mom5(MakefilePackage):
...
+    version("master", branch="master", preferred=True)
+    version("access-esm1.5", branch="access-esm1.5")
 
-    variant("deterministic", default=False, description="Deterministic build.")
     variant("type", default="ACCESS-OM", description="Build MOM5 to support a particular use case.", values=("ACCESS-CM", "ACCESS-ESM", "ACCESS-OM", "ACCESS-OM-BGC", "MOM_solo"), multi=False)
     variant("restart_repro", default=True, description="Reproducible restart build.")
-    variant("optimisation_report", default=False, description="Generate optimisation reports.")
+    with when("@master"):
+        variant("deterministic", default=False, description="Deterministic build.")
+        variant("optimisation_report", default=False, description="Generate optimisation reports.")
```
harshula, line 26: Does this work when `@git.<tag>` is used?

packages/mom5/package.py
```diff
@@ -14,24 +14,31 @@ class Mom5(MakefilePackage):
...
     # Depend on virtual package "mpi".
     depends_on("mpi")
     depends_on("netcdf-fortran@4.5.2:")
     depends_on("netcdf-c@4.7.4:")
-    depends_on("datetime-fortran")
-    depends_on("oasis3-mct+deterministic", when="+deterministic")
-    depends_on("oasis3-mct~deterministic", when="~deterministic")
-    depends_on("libaccessom2+deterministic", when="+deterministic")
-    depends_on("libaccessom2~deterministic", when="~deterministic")
+    with when("@master"):
```
harshula, line 34: Try:
```suggestion
    with not when("@access-esm1.5"):
```

packages/mom5/package.py
```diff
@@ -312,33 +475,45 @@ def build(self, spec, prefix):
...
+                # requires the --no_version switch to avoid git hashes being
+                # embedded in the binary.
+                build(
+                    "--type",
+                    self.spec.variants["type"].value,
+                    "--platform",
+                    self._platform,
+                    "--no_environ",
+                    "--no_version"
+                )
```
harshula, line 504:
```suggestion
            if not "@access-esm1.5" in self.spec:
                build.add_default_arg("--no_version")
            if "+optimisation_report" in self.spec:
                build.add_default_env("REPORT", "true")
            if "+restart_repro" in self.spec:
                build.add_default_env("REPRO", "true")
            # The MOM5 commit d7ba13a3f364ce130b6ad0ba813f01832cada7a2
            # requires the --no_version switch to avoid git hashes being
            # embedded in the binary.
            build(
                "--type",
                self.spec.variants["type"].value,
                "--platform",
                self._platform,
                "--no_environ",
            )
```

harshula (approved):
Looks good, well done!

PR #107, opened 2024-05-15
access-esm1p5: add dependency mom5@access-esm1.5

harshula:
Another component is ready for testing.


CodeGat (approved):
Looks like it's coming together well. I should have a `spack.yaml` ready at around the same time. Is the `um7` the final dependency?

harshula:
Thanks! `um7` SPD is ready and apparently works. I'll wait till the infrastructure is ready before enabling it.

PR #118, opened 2024-06-05
Add version access-esm1.5 to um7

### LINKED ISSUE #117: Add version access-esm1.5 to um7
Add version `access-esm1.5` to [the `um7` Spack package](https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/um7/package.py), using the [access-esm1.5 branch of UM_v7](https://github.com/ACCESS-NRI/UM_v7/tree/access-esm1.5). This is for consistency with the other `access-esm1.5`-related spack builds, and allows for other versions in future.

penguian:
Closes #117.

Tested by successfully running
```
spack install um7@access-esm1.5^openmpi@4.0.2%intel@19.0.5.281 arch=linux-rocky8-x86_64
```

### DIFF
--- a/packages/um7/package.py
+++ b/packages/um7/package.py
@@ -24,6 +24,7 @@ class Um7(Package):
 
     # https://code.metoffice.gov.uk/trac/um/wiki/PastReleases
     version("7.3")
+    version("access-esm1.5", branch="access-esm1.5")
 
     maintainers("penguian")
 
--- a/packages/um7/package.py
+++ b/packages/um7/package.py
@@ -23,7 +23,7 @@ class Um7(Package):
     git = "git@github.com:ACCESS-NRI/UM_v7.git"
 
     # https://code.metoffice.gov.uk/trac/um/wiki/PastReleases
-    version("7.3")
+    version("7.3", branch="main", preferred=True)
     version("access-esm1.5", branch="access-esm1.5")
 
     maintainers("penguian")


packages/um7/package.py
```diff
@@ -24,6 +24,7 @@ class Um7(Package):
 
     # https://code.metoffice.gov.uk/trac/um/wiki/PastReleases
     version("7.3")
+    version("access-esm1.5", branch="access-esm1.5")
```
harshula, line 27: Do you want to use `preferred=True` like you did in the MOM5 SPD for one of the `version()` calls?

penguian: I am not sure. I don't think that it is needed until we add "access-esm1.6". 
If I did, add `preferred=True`, it would be
```
version("7.3", branch="main", preferred=True)
```
and I don't think that would be useful until ACCESS-ESM1.6 changes are merged into `main` as per https://github.com/ACCESS-NRI/UM_v7/issues/8

What do you think?

harshula: Since you found `preferred=True`, my preference is to have one version set as preferred when there are multiple versions and some are non-numeric.

penguian: Will do.

penguian: Done.

harshula (approved):

penguian:
Merging.

PR #127, opened 2024-07-05
access-om2: pass type to depends_on()

harshula:
* Passing type="run" to help with module autoloading as per: https://spack.readthedocs.io/en/v0.22.0/module_file_support.html#autoloading-and-hiding-dependencies


harshula:
There was a bug in the code that was reviewed. It was an exact copy of the access-om2-bgc SBD and I had not removed  `type=ACCESS-OM-BGC`. I noticed the bug in the pre-release build and force pushed the fix so that we could start another pre-release build quickly.

PR #128, opened 2024-07-08
Make um7 Spack package more consistent with other ACCESS-NRI Spack packages

### LINKED ISSUE #122: Make um7 path definitions more consistent with other ACCESS-NRI packages
The current path definitions for `um7` use:

- `PATH` to locate `fcm`.
- `env.prepend_path("CPATH", ...)` to define paths to headers and Fortran modules. 
- `env.prepend_path("LIBRARY_PATH", ...)` to define paths to libraries at build time.
- `env.prepend_path("LD_LIBRARY_PATH", ...)` to define the path to `netcdf-fortran` at runtime.

Similarly to `cice4` and other ACCESS-NRI Spack package definitions:
- Using `depends_on("fcm", ...)` would be preferable, since it allows for versioning.
- Paths to headers and Fortran modules would be better defined similar to the following from `cice4`.
```
istr = " ".join([
                join_path((spec["oasis3-mct"].headers).cpp_flags, "psmile.MPI1"),
                join_path((spec["oasis3-mct"].headers).cpp_flags, "mct")])
ideps = ["netcdf-fortran",...]
incs = " ".join([istr] + [(spec[d].headers).cpp_flags for d in ideps])
```
- Library paths should be specified via `RPATH`, similar to the following from `cice4`.
```
    def get_linker_args(self, spec, name):
        return " ".join(
                    [(spec[name].libs).ld_flags,
                    "-Wl,-rpath=" + join_path(spec[name].prefix, "lib")]
                   )
...
        lstr = ""
        ldeps = ["oasis3-mct", "netcdf-fortran"]
        libs = " ".join([lstr] + [self.get_linker_args(spec, d) for d in ldeps])
```

See also #113 for another `um7`-related issue.

penguian:
Closes #122 
The changes are extensive, but should not affect the default or `debug` use cases.

1. Remove the unused `"omp"` and `"netcdf"` variants.
2. Change version range for `depends_on("netcdf-fortran")` from `4.5.1:4.5.2` to just `4.5.2`.
3. Use `edit`, `build` and `install` phases.
   - In the `edit` phase, create an FCM config file whose contents depend on `opt_value`.
   - In the `build` phase, use FCM and the Intel compiler to build the executable.
   - In the `install` phase, install the executable into the `prefix.bin` directory.
4. In `setup_build_environment`, use list comprehensions to define `oasis3_incs` and `incs`, and use a loop to prepend to `CPATH` for each included directory.
5. Prepend to `LIBRARY_PATH` only for `gcom4`, since that library is the only one that must be statically linked.
6. Remove `setup_run_environment()` : `edit()` now uses `_get_linker_args()` which uses `-rpath`. 
7. Remove the dependence of `bld_cfg_path()` on `opt_value`. The `edit` phase now creates the FCM config file.
8. In the `edit` phase, do not define the unused environment variables `"optimisation_level"`, `"openmp"`, and `"netcdf"`

### DIFF
--- a/packages/um7/package.py
+++ b/packages/um7/package.py
@@ -10,9 +10,6 @@
 
 from spack.package import *
 
-import os
-import os.path
-
 class Um7(Package):
     """
     UM is a numerical weather prediction and climate modelling software package.
--- a/packages/um7/package.py
+++ b/packages/um7/package.py
@@ -32,7 +32,7 @@ class Um7(Package):
     depends_on("netcdf-fortran@4.5.2", type=("build", "link"))
     depends_on("oasis3-mct@access-esm1.5", type=("build", "link"))
 
-    variant("opt", default="high", description="Optimization level",
+    variant("optim", default="high", description="Optimization level",
             values=("high", "debug"), multi=False)
 
     phases = ["edit", "build", "install"]
@@ -55,24 +55,18 @@ class Um7(Package):
         env.prepend_path("LIBRARY_PATH", self.spec["gcom4"].prefix.lib)
 
 
-    def _bld_path(self):
-        """
-        Return the path to the build directory.
-        """
-        return "ummodel_hg3"
+    # The path to the build directory.
+    _bld_path = "ummodel_hg3"
 
-    def _bld_cfg_path(self):
-        """
-        Return the path to the build configuration.
-        """
-        return join_path(self._bld_path(), "cfg", "bld-hadgem3-spack.cfg")
+    # The path to the build configuration.
+    _bld_cfg_path = join_path(_bld_path, "cfg", "bld-hadgem3-spack.cfg")
 
 
-    def _exe_name(self, opt_value):
+    def _exe_name(self, optim_value):
         """
-        Return the executable name, depending on opt_value.
+        Return the executable name, depending on optim_value.
         """
-        if opt_value == "debug":
+        if optim_value == "debug":
             return "um_hg3_dbg.exe"
         else:
             return "um_hg3.exe"
@@ -99,7 +93,7 @@ class Um7(Package):
         ldeps = ["oasis3-mct", "netcdf-fortran", "dummygrib"]
         libs = " ".join([self._get_linker_args(spec, d) for d in ldeps] + ["-lgcom"])
 
-        opt_value = spec.variants["opt"].value
+        opt_value = spec.variants["optim"].value
         EXE_NAME = self._exe_name(opt_value)
         CPPKEYS = (
             "C_LONG_LONG_INT=c_long_long_int MPP=mpp C_LOW_U=c_low_u "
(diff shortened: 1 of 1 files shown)


penguian:
The output of the `pre-industrial` config from `ACCESS-NRI/access-esm1.5-configs` reproduces bitwise.
See `/g/data/tm70/pcl851/src/ACCESS-NRI/access-esm1.5-configs` Git branch `pre-industrial-84-um7-modules` and output in `/scratch/tm70/pcl851/access-esm/archive/access-esm1.5-configs-pre-industrial-84-um7-modules-1be4de64` for details.

packages/um7/package.py
```diff
@@ -28,38 +25,34 @@ class Um7(Package):
...
         env.prepend_path("LIBRARY_PATH", self.spec["gcom4"].prefix.lib)
-        env.prepend_path("LIBRARY_PATH", self.spec["netcdf-fortran"].prefix.lib)
-        env.prepend_path("LIBRARY_PATH", self.spec["oasis3-mct"].prefix.lib)
-
-
-    def setup_run_environment(self, env):
-        env.prepend_path("LD_LIBRARY_PATH", self.spec["netcdf-fortran"].prefix.lib)
 
 
     def _bld_path(self):
```
harshula, line 58: Can this be a member variable instead of a member function?

harshula, line 58: Can this be a member variable instead of a member function?

packages/um7/package.py
```diff
@@ -68,16 +61,11 @@ def _bld_path(self):
         """
         return "ummodel_hg3"
 
-
-    def _bld_cfg_path(self, opt_value):
+    def _bld_cfg_path(self):
```
harshula, line 64: Can this be a member variable instead of a member function?

packages/um7/package.py
```diff
@@ -28,38 +25,34 @@ class Um7(Package):
...
+    depends_on("dummygrib", type=("build", "link"))
     depends_on("gcom4@access-esm1.5+mpi", type=("build", "link"))
     depends_on("openmpi@4.0.2:4.1.0", type=("build", "run"))
-    depends_on("netcdf-fortran@4.5.1:4.5.2", type=("build", "link"))
+    depends_on("netcdf-fortran@4.5.2", type=("build", "link"))
     depends_on("oasis3-mct@access-esm1.5", type=("build", "link"))
 
-    variant("omp", default=True, description="Use OpenMP")
-    variant("netcdf", default=True, description="NetCDF")
     variant("opt", default="high", description="Optimization level",
```
harshula, line 35: Can variant "opt" be renamed to something that indicates 'optimisation'?

packages/um7/package.py
```diff
@@ -90,48 +78,147 @@ def _exe_name(self, opt_value):
...
+        with open(bld_cfg_path, "w") as bld_cfg_file:
+            bld_cfg_file.write(config)
 
+
+    def build(self, spec, prefix):
+        """
+        Use FCM and the Intel compiler to build the executable.
+        """
         if self.compiler.name == "intel":
             env["platform_config_dir"] = "nci-x86-ifort"
```
harshula, line 207: Now that we create the config file from the SPD, can we define config dir whatever we like?

packages/um7/package.py
penguian: Done.

packages/um7/package.py
penguian: Done.

packages/um7/package.py
penguian: Changed to `"optim"`.

packages/um7/package.py
penguian: Removed the check since the environment variable is not used.

harshula (approved):
Reviewed in VC. Assuming the configuration in the variable `config` is correct.

penguian:
@harshula  Thanks for the approval. I will resolve all conversations.
You can check the configuration against https://github.com/ACCESS-NRI/UM_v7/blob/access-esm1.5/ummodel_hg3/cfg/bld-hadgem3-mct.cfg and https://github.com/ACCESS-NRI/UM_v7/blob/access-esm1.5/ummodel_hg3/cfg/bld-dbg-hadgem3-C2.cfg if you want.

PR #139, opened 2024-08-12
Configure UM via models

### LINKED ISSUE #135: Configure UM via models
Make the UM Spack package more user configurable by adding variants corresponding to the variables in [packages/um/rose-app.conf](https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/um/rose-app.conf). 

Create a `model` subdirectory and a `model` variant, and move the existing [packages/um/rose-app.conf](https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/um/rose-app.conf) to [packages/um/model/access-ram3/rose-app.conf](https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/um/model/access-ram3/rose-app.conf). The `model` variant is to be used to obtain default values for the other variants. 

The exception is the UM version itself, which must be configured as usual, with a check that the UM version and the model are compatible.

penguian:
Closes #135 

The changes are extensive.

1. The first main change is the introduction of the `model` variant, and a corresponding `model/` directory. The string value of `model` determines which subdirectory of `model/` to use to find the corresponding `rose-app.conf` file. There are currently 3 subdirectories, `vn13` (default), `vn13p0-rns` (corresponding to the RNS suite `u-dg768`) and `vn13p5-rns`.
2. The second main change is that almost every environment variable in `model/*/rose-app.conf` now has a corresponding variant.
   1. The environment variables in `rose-app.conf` that take the values `true` or `false` correspond to either `bool` or off/on variants. 
      1. The `bool` variants are normal Spack `bool` variants, taking the value `False` or `True`. The lambda function`bool_to_str` transforms these to the environment variable values `"false"` or `"true"`, determined by the value of the variant. The value given in `rose-app.conf` is ignored.  Currently only `eccodes` and `netcdf` are `bool` variants. The default value of  these`bool` variables is `True`, corresponding to the value used in the RNS suite `u-dg768`.
      2. The off/on variants take the values `"none"`, `"off"` or `"on"`. The Spack default is `"none"`. Leaving the value as `"none"` results in the corresponding environment variable being set to the value given in `rose-app.conf`. Otherwise, the lambda function `off_on_to_str` transforms `"off"` to `"false"` or `"on"` to `"true"` and this is the value used for the corresponding environment variable, overriding the value given in `rose-app.conf`. 
   3. The component revision environment variables in `rose-app.conf`, other than `um_rev`, correspond to revision variants. For revision variants, the Spack default is `"none"`. If the variant value left as `"none"`, then if the value in `rose-app.conf` defined and non-empty, it is used as the environment variable value, otherwise the value is `"um{version}"` where `version` is the UM version. If the variant value is not `"none"` then it is used as the environment variable value, overriding any value in `rose-app.conf`.
   4. All other environment variables in `rose-app.conf`, with the exception of `"um_rev"`, correspond to `str` variants. The default value for each of these variants is `"none"`. Leaving this value as `"none"` results in the corresponding environment variable being set to the value given in `rose-app.conf`. If the variant is set to a value other than `"none"`, then this value overrides the value given in `rose-app.conf`. This effectively makes the value in `rose-app.conf` the default.


penguian:
All comments addressed, except that `eccodes` and `netcdf` are boolean variants so that dependencies can be specified. These both default to `True`.

packages/um/package.py
```diff
@@ -83,36 +166,108 @@ def setup_build_environment(self, env):
...
+        # Override those environment variables where an off/on variant is specified.
+        off_on_to_str = lambda off_on: "true" if off_on == "on" else "false"
+        for var in self._off_on_variants:
+            spec_value = spec.variants[var].value
+            if spec_value != "none":
+                spec_str_value = off_on_to_str(spec_value)
+                check_model_vs_spec(model, config_env, var, spec_str_value)
+                config_env[var] = spec_str_value
+
+        # Overide those environment variables where a revision variant is specified.
```
harshula, line 241: s/Overide/Override/

harshula, line 241: s/Overide/Override/

packages/um/package.py
```diff
@@ -83,36 +166,108 @@ def setup_build_environment(self, env):
...
+            spec_value = spec.variants[var].value
+            if spec_value != "none":
+                check_model_vs_spec(model, config_env, var, spec_value)
+                config_env[var] = spec_value
+
+        # Get the linker arguments for some dependencies.
         for fcm_libname in ["eccodes", "netcdf"]:
             linker_args = self._get_linker_args(spec, fcm_libname)
-            env.set(f"ldflags_{fcm_libname}_on", linker_args)
+            config_env[f"ldflags_{fcm_libname}_on"] = linker_args
```
harshula, line 262: Could be done with one assignment.

harshula (approved):
Looks good, well done!

packages/um/package.py
penguian: Done.

packages/um/package.py
penguian: I left this as-is, as it is easy enough to understand what is happening here without combining the two lines into one assignment. Also, a single assignment would result in a 94 character line, which would be the longest line in the file, unless the line was somehow wrapped, continued or shortened.

harshula (approved):

PR #141, opened 2024-08-20
ACCESS-ESM1.6 SBD

### LINKED ISSUE #140: Add ACCESS-ESM1.6 SBD
I would like to add an initial SBD for ACCESS-ESM1.6. For now, the only difference from the ACCESS-ESM1.5 SBD will be the MOM5 spec.

In ACCESS-ESM1.6, I would like to build MOM5 in the same way as ACCESS-ESM1.5, except **using the "master" branch and the "ACCESS-ESM" "type" variant**. To allow this, the following changes are required:

In MOM5 SPD:
- Add an "access-esm1.6" version that uses the "master" branch
- Remove the [`conditional` logic](https://github.com/ACCESS-NRI/spack-packages/blob/3f45685e8f4b2e10d03db5bcd0027bc22df475a4/packages/mom5/package.py#L32) from the definition of the "type" variant
- Extend `with`s and `if`s currently set up for `@access-esm1.5` to include `@access-esm1.6`

In ACCESS-ESM1.5 SPD:
- Specify `+type=ACCESS-CM` in MOM5 spec

In (new) ACCESS-ESM1.6 SPD:
- Specify `+type=ACCESS-ESM` in MOM5 spec

I'll follow up with a PR shortly.

dougiesquire:
This PR adds a new SBD for ACCESS-ESM1.6. At the moment, this is only differs from ACCESS-ESM1.5 in how MOM5 is built.

Closes #140

### DIFF
--- a/packages/mom5/package.py
+++ b/packages/mom5/package.py
@@ -40,7 +40,7 @@ class Mom5(MakefilePackage):
         depends_on("oasis3-mct~deterministic", when="~deterministic")
         depends_on("libaccessom2+deterministic", when="+deterministic")
         depends_on("libaccessom2~deterministic", when="~deterministic")
-    with when("@access-esm1:access-esm1.6"):
+    with when("@access-esm1.5:access-esm1.6"):
         depends_on("netcdf-c@4.7.1:4.7.4")
         depends_on("netcdf-fortran@4.5.1:4.5.2")
         # Depend on "openmpi".
@@ -61,7 +61,7 @@ class Mom5(MakefilePackage):
         config = {}
 
         # NOTE: The order of the libraries matters during the linking step!
-        if ("@access-esm1.5" in self.spec) or ("@access-esm1.6" in self.spec):
+        if self.spec.satisfies("@access-esm1.5:access-esm1.6"):
             istr = " ".join([
                     join_path((spec["oasis3-mct"].headers).cpp_flags, "psmile.MPI1"),
                     join_path((spec["oasis3-mct"].headers).cpp_flags, "mct")])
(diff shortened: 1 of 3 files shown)


packages/mom5/package.py
```diff
@@ -490,7 +487,7 @@ def build(self, spec, prefix):
             build = Executable("./MOM_compile.csh")
             if "+restart_repro" in self.spec:
                 build.add_default_env("REPRO", "true")
-            if "@access-esm1.5" not in self.spec:
+            if self.spec.satisfies("@access-esm1.5"):
```
penguian, line 490: Should this be
```
if not self.spec.satisfies("@access-esm1.5:access-esm1.6"):
```
because of the `not` on line 493 of the previous?

penguian (changes requested):
See my comment on line 490.

packages/mom5/package.py
dougiesquire: Whoops, good catch. I was ~not~ paying enough attention when I changed the syntax. I think it should be

```
if not self.spec.satisfies("@access-esm1.5")
```

because the `master` and `access-esm1.6` versions both use the master branch and should enter here.

I'm away from my computer at the mo, but will change this when I'm back.

dougiesquire:
I think I need approval from @penguian to allow the merge

CodeGat:
Admins can dismiss requested changes, if you think he would be happy - just since he's not in this morning

dougiesquire:
I've resolved his review comment, so I think he'd be happy?

CodeGat:
Done @dougiesquire

PR #149, opened 2024-09-20
access-test: new SPD to test ACCESS-NRI infrastructure

harshula:
@atteggiani made a good point that ACCESS-OM2 takes too long to build for testing our Spack instance. That got me thinking whether a Test/QA fake model release would be useful for CI, CD, Spack, Docker and Gadi testing ...

@CodeGat: "I'm starting to think more and more that it'd be a great idea."


aidanheerdegen (approved):
LGTM

PR #153, opened 2024-10-14
Added specific reference to 'master' version in mom5 and oasis packages

### LINKED ISSUE #152: Set dependencies conditions to specific defined versions
### Description of the problem
Currently, the conditions for the dependencies of the [mom5 package](https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/mom5/package.py) are defined in a way that does not explicitly mention a [defined version](https://github.com/ACCESS-NRI/spack-packages/blob/a49002968d99b0161ba57be15a0dba0f76bdbdbc/packages/mom5/package.py#L20-L22):

https://github.com/ACCESS-NRI/spack-packages/blob/a49002968d99b0161ba57be15a0dba0f76bdbdbc/packages/mom5/package.py#L33

This, according to [Spack version comparison documentation](https://spack.readthedocs.io/en/latest/packaging_guide.html#version-comparison), would include the defined version `master`. However, at a first glance, it is not too obvious and is prone to errors.

### Proposed solution
I suggest setting the dependencies based on specific defined versions, for clarity and simplicity.

In this specific case, the dependencies could be set as:
```diff
- with when("@:access-esm0,access-esm2:"):
+ with when("@master"):   
    depends_on("netcdf-c@4.7.4:")
    depends_on("netcdf-fortran@4.5.2:")
    # Depend on virtual package "mpi".
    depends_on("mpi")
    depends_on("datetime-fortran")
    depends_on("oasis3-mct+deterministic", when="+deterministic")
    depends_on("oasis3-mct~deterministic", when="~deterministic")
    depends_on("libaccessom2+deterministic", when="+deterministic")
    depends_on("libaccessom2~deterministic", when="~deterministic")
with when("@access-esm1.5:access-esm1.6"):
    depends_on("netcdf-c@4.7.1:4.7.4")
    depends_on("netcdf-fortran@4.5.1:4.5.2")
    # Depend on "openmpi".
    depends_on("openmpi@4.0.2:4.1.0")
    depends_on("oasis3-mct@access-esm1.5")
```

`access-esm1.5` and `access-esm1.6` versions could be further separated if different dependencies are required.

This would likely not add any overhead when new versions need to be added, because the `package.py` file would still need to be changed and the dependencies of the new version checked. In that case, a further condition can be added for the new version (or the new version can be included in an already-present condition).

This would improve the package clarity.

atteggiani:
- [x] Fixes #152.
- [x] Fixed "f-string withoug placeholders" issue with `Ruff` in oassis.

### DIFF
--- a/packages/mom5/package.py
+++ b/packages/mom5/package.py
@@ -30,7 +30,7 @@ class Mom5(MakefilePackage):
         multi=False,
         description="Build MOM5 to support a particular use case.")
 
-    with when("@:access-esm0,access-esm2:"):
+    with when("@master"):
         depends_on("netcdf-c@4.7.4:")
         depends_on("netcdf-fortran@4.5.2:")
         # Depend on virtual package "mpi".
--- a/packages/oasis3-mct/package.py
+++ b/packages/oasis3-mct/package.py
@@ -23,7 +23,7 @@ class Oasis3Mct(MakefilePackage):
     variant("deterministic", default=False, description="Deterministic build.")
     variant("optimisation_report", default=False, description="Generate optimisation reports.")
 
-    with when("@:access-esm0,access-esm2:"):
+    with when("@master"):
         depends_on("netcdf-fortran@4.5.2:")
         # Depend on virtual package "mpi".
         depends_on("mpi")
--- a/packages/oasis3-mct/package.py
+++ b/packages/oasis3-mct/package.py
@@ -251,7 +251,7 @@ f90         = $(F90)
 f           = $(F90)
 """
 
-        config["gcc"] = f"""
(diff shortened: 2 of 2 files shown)


penguian (commented):
Changes seem fine to me, but please ask @harshula to also review. I think that he might want to leave the 'f' strings as-is.

harshula:
Hi @atteggiani , Please leave the 'f' strings as-is. I need the to modify the compiler flags more dynamically in the future.

harshula:
Sorry, just remembered why we shouldn't do this change. We need the first `with when()` to work with the `development` branch.

harshula:
NOTABUG

PR #155, opened 2024-10-22
Update package.py for GCOM package

anton-seaice:
Switch to https for git to avoid need for ssh key password. (Maybe there is a reason for this ?)


harshula:
Hi @anton-seaice , I've assigned the review to @penguian . I assume this has something to do with the licensing of the software.

penguian (approved):
Appoved.

penguian:
Sorry, looks like I got this wrong. "git@github.com:ACCESS-NRI/GCOM4.git" is a private GitHub repository and as such needs a login. The simplest way to log in is to use SSH.

PR #157, opened 2024-11-05
access-mocsy: create an SPD for ACCESS-NRI's mocsy fork

harshula:


penguian (approved):
Is it unusual that the Makefile does not list module files?

harshula:
> Is it unusual that the Makefile does not list module files?

The other example I've seen is oasis3-mct's Makefile. That doesn't list the module files either.

penguian (commented):
Approved.

PR #169, opened 2024-11-21
mom5: use a more accurate name for an environment variable

harshula:
While debugging an issue, I found that the variable name `$SPACK_BUILD` confusing because we use Spack to build MOM5 regardless of whether that variable is set to true of false.

### DIFF
--- a/packages/mom5/package.py
+++ b/packages/mom5/package.py
@@ -498,7 +498,7 @@ TMPFILES = .*.m *.T *.TT *.hpm *.i *.lst *.proc *.s
                 build.add_default_env("REPRO", "true")
 
             if self.spec.satisfies("+access-gtracers"):
-                build.add_default_env("SPACK_BUILD", "true")
+                build.add_default_env("SPACK_FMS_EXTERNAL", "true")
 
             if not self.spec.satisfies("@access-esm1.5"):
                 # The MOM5 commit d7ba13a3f364ce130b6ad0ba813f01832cada7a2


harshula:
Before merging this to `main`, I have to commit the following to MOM5's `development` branch:
```
diff --git a/exp/MOM_compile.csh b/exp/MOM_compile.csh
index b095b22..272054a 100755
--- a/exp/MOM_compile.csh
+++ b/exp/MOM_compile.csh
@@ -156,7 +156,7 @@ if ( $cosima_version ) then
     set cppDefs = "$cppDefs -DCOSIMA_VERSION"
 endif
 
-if ( ! $?SPACK_BUILD ) then
+if ( ! $?SPACK_FMS_EXTERNAL ) then
     echo "Building type=$type with internal FMS"
     source ./FMS_compile.csh
     set includes = "-I$code_dir/shared/include -I$executable:h:h/lib_FMS -I$executable:h:h/lib_ocean"
@@ -264,7 +264,7 @@ if( $type == ACCESS-OM || $type == ACCESS-ESM) then
     set srcList = ( $srcList access/shared )
 endif
 
-if ( ! $?SPACK_BUILD ) then
+if ( ! $?SPACK_FMS_EXTERNAL ) then
     set libs = "$libs $executable:h:h/lib_FMS/lib_FMS.a"
     if( $type == ACCESS-CM || $type == ACCESS-ESM || $type == ACCESS-OM || $type == ACCESS-OM-BGC) then
         set includes = "$includes -I$executable:h:h/lib_FMS"
```

aidanheerdegen (approved):
LGTM.

Issue #163, opened 2024-11-14
Make spack packages for CICE6, MOM6, WWIII, access3-share


anton-seaice:
Split the existing access-om3-nuopc package into several components, to produce the same result as the access-om3-nuopc package, split into components.

anton-seaice:
We want to be able to build both the MOM6/CICE6 standalone and access builds.

According to the spack docs, it looks like you can only use git hashes/tags from one repository - see https://spack.readthedocs.io/en/latest/packaging_guide.html#fetching-from-code-repositories

I think this means that we would have one spack package for each component, which points to the access-nri fork, and a seperate spack package (contributed to the upstream spack repo) for the upstream build ?

Should we prefix out packages with _access-_ so we can have different packages for the upstream version in the future. 

i.e  there is an _access-cice_ package in this repository and a _cice_ package in https://github.com/spack/spack/tree/develop/var/spack/repos/builtin/packages ?

This seems like duplication, needing two spack packages for each model component ...

anton-seaice:
As part of this work, I am making a _access3-exe_ package.

The code to produce an executable for access-om3 and access-cm3 are the same ( https://github.com/ESCOMP/CMEPS/blob/main/cesm/driver/esmApp.F90 ), it just needs to be linked to different libraries.

currently a lib is made like [this](https://github.com/ACCESS-NRI/dev-access-cm3-build/blob/main/build.sh):

`
# make UM library
object_files=""
for file in $ROSE_SUITE_DIR/share/fcm_make/build-atmos/o/*; do
    if [[ $file == *"driver.o" ]]; then
        echo $file
    elif [[ $file == *"um_main.o" ]]; then
        echo $file
    else
        object_files="${object_files} ${file}"
    fi
done

ar -r libum.a $object_files
`

and linked in through CMake . We should make sure the build is generic to access-om3/cm3/esm3 and avoid using 'om3' as a specific label if possible. ( e.g. in https://github.com/ACCESS-NRI/access3-share/pull/2, change OM3 to Access3). Ill plan to leave the adding of UM to a later issue.

q1165600785:
Have you finished making a separate mom6 package? I am using autoreconf to build, but after the build is successful, it always fails to run. I am looking forward to using spack to install mom6 without installing the dependencies required by mom6. Thanks for your hard work.

aidanheerdegen:
> According to the spack docs, it looks like you can only use git hashes/tags from one repository

[...]

PR #171, opened 2024-11-27
Spack Components for OM3

anton-seaice:
This is an extremely draft collection of changes to `spack-packages` to build OM3 by component.

I've been building access3-share with '+install_libraries' in the spec, I haven't looked at building with an executable yet

I've been using these git versions :

https://github.com/ACCESS-NRI/CICE/tree/5-port_cmake_build

and 

https://github.com/ACCESS-NRI/access3-share/tree/share_prototype

Example deployment: https://github.com/ACCESS-NRI/ACCESS-OM3/pull/35


anton-seaice:
Contributes to #163

anton-seaice:
This is still a draft but is ready for a review for style / approach.

I've only implemented for CICE6 at this point, see the example deployment in https://github.com/ACCESS-NRI/ACCESS-OM3/pull/35

anton-seaice:
In deployment, I have been trying to build 'access3-exe' with 'configurations=CICE6'

When concretising, the requested configuration gets [added](https://github.com/ACCESS-NRI/ACCESS-OM3/actions/runs/12761097471/job/35567599971#step:9:796) to the default:

```
 -   myn2njk      ^access3-exe@git.share_prototype=0-git.175%intel@2021.10.0~ipo~openmp build_system=cmake build_type=Release configurations=CICE6,CICE6-WW3,MOM6-CICE6,MOM6-CICE6-WW3 generator=make arch=linux-rocky8-x86_64
```

The default in the access3-exe spack package is for 'configurations = CICE6-WW3,MOM6-CICE6,MOM6-CICE6-WW3' and so the CICE6 gets added to this list rather than replacing it. There doesn't appear to be a way in spack to remove variants, when `multi=True` (i.e. multiple are supported). So i think I will set the default to be `None`.

The spack _MultiValuedVariant_ class only has an append function (no remove function ...), see https://github.com/spack/spack/blob/2447d16e55d800e6a9cb82a645ab9edfd24f08ac/lib/spack/spack/variant.py#L503

packages/mom6/package.py
```diff
@@ -0,0 +1,73 @@
...
+
+    root_cmakelists_dir = "cmake"
+
+    variant(
+        "install_libraries",
+        default=False,
+        description="Install component libraries"
+    )
+    variant("openmp", default=False, description="Enable OpenMP")
+    variant("mom_symmetric", default=False, description="Use symmetric memory in MOM6")
```
anton-seaice, line 48: We probably make mom_symmetric True by default, need to confirm. See https://github.com/ACCESS-NRI/ACCESS-OM3/pull/35

packages/access3-exe/package.py
```diff
@@ -0,0 +1,88 @@
...
+        multi=True,
+        description="ACCESS-OM3 configurations to build",
+    )
+    
+    variant("openmp", default=False, description="Enable OpenMP")
+    
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("access-cice+cesmcoupled+access3")
+    depends_on("access-mom6+cesmcoupled+access3")
```
anton-seaice, line 51: Should these have when statements ?

packages/access-mom6/package.py
```diff
@@ -0,0 +1,62 @@
...
+    git = "https://github.com/ACCESS-NRI/MOM6.git"
+    submodules = True
+    url = "https://github.com/ACCESS-NRI/MOM6.git"
+
+    version("cmake_build",
+            git="https://github.com/ACCESS-NRI/MOM6.git",
+            branch="cmake_build",
+            submodules=True,
+            preferred=True,
+            )
```
anton-seaice, line 26:
```suggestion
    version("dev/access", branch="dev/access",preferred=True)
```

packages/access-mom6/package.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+    root_cmakelists_dir = "cmake"
+
+    def cmake_args(self):
+        args = [
+            self.define_from_variant("OPENMP", "openmp"),
+            self.define_from_variant("ENABLE_MOM_SYMMETRIC", "mom_symmetric"),
+            self.define_from_variant("ENABLE_ACCESS_MOM6", "access3"),
+            self.define_from_variant("ENABLE_CESMCOUPLED", "cesmcoupled"),
```
anton-seaice, line 55:
```suggestion
            self.define_from_variant("ENABLE_MOM_SYMMETRIC", "mom_symmetric"),
            self.define_from_variant("ACCESS3_MOM6", "access3"),
            self.define_from_variant("CESMCOUPLED", "cesmcoupled"),
```

packages/access-cice/package.py
```diff
@@ -0,0 +1,66 @@
...
+    depends_on("access3-share", when="+access3") 
+    # depends_on("access3-share+openmp", when="+openmp+access3")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:", when="io_type=NetCDF")
+    depends_on("parallelio@2.5.10: build_type==RelWithDebInfo", when="io_type=PIO")
+    depends_on("parallelio fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model source' cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'", when="%intel io_type=PIO")
+
+    root_cmakelists_dir = "configurations/scripts/cmake"
```
anton-seaice, line 51:
```suggestion
    root_cmakelists_dir = "configuration/scripts/cmake"
```

packages/access-cice/package.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class AccessCice(CMakePackage):
+    """CICE is a computationally efficient model for simulating the growth, melting, 
+    and movement of polar sea ice. CICE is maintained and developed by the CICE-Consortium. 
+    This package builds using the Access3Share common libraries for ACCESS 3 models."""
+
+    homepage = "https://github.com/CICE-Consortium/CICE"
+    url = "https://github.com/CICE-Consortium/CICE/archive/refs/tags/CICE6.6.0.tar.gz"
+    version("6.6.0", md5="1c678c0af67bf09f92c0a861344c3a92")
```
anton-seaice, line 17:
```suggestion
```

Shall we pre-emptively create a version("access-om3", branch="main") here ?

packages/access3-exe/package.py
```diff
@@ -0,0 +1,73 @@
...
+    for configurations in ["MOM6", "MOM6-CICE6", "MOM6-CICE6-WW3", "MOM6-WW3"]:
+        depends_on("access-mom6+cesmcoupled+access3")
+    # for configurations in ["WW3", "CICE6-WW3", "MOM6-CICE6-WW3", "MOM6-WW3"]:
+        # depends_on("access-ww3 driver=nuopc/cmeps +cesmcoupled")
+    
+    flag_handler = CMakePackage.build_system_flags
+
+    def cmake_args(self):
+
+        #make configurations a cmake argument
```
anton-seaice, line 62:
```suggestion
        # make configurations a cmake argument
```

packages/access3-exe/package.py
```diff
@@ -0,0 +1,73 @@
...
+    license("Apache-2.0", checked_by="anton-seaice")
+
+    variant(
+        "build_type",
+        default="Release",
+        description="The build type to build",
+        values=("Debug", "Release"),
+    )
+
+    # To-DO: confirm if we want a MOM6 only for regional modelling
```
anton-seaice, line 28:
```suggestion
```

packages/access-cice/package.py
aidanheerdegen: How does this interact with the `access3` variant?

packages/access-cice/package.py
```diff
@@ -0,0 +1,66 @@
...
+        values=("Debug", "Release"),
+    )
+
+    depends_on("access3-share", when="+access3") 
+    # depends_on("access3-share+openmp", when="+openmp+access3")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:", when="io_type=NetCDF")
+    depends_on("parallelio@2.5.10: build_type==RelWithDebInfo", when="io_type=PIO")
```
aidanheerdegen, line 48: This is overriding the global `build_type` variant. Is that intentional?

packages/access3-share/package.py
```diff
@@ -0,0 +1,55 @@
...
+
+    license("Apache-2.0", checked_by="anton-seaice")
+
+    # variant("openmp", default=False, description="Enable OpenMP")
+
+    variant(
+        "build_type",
+        default="Release",
+        description="The build type to build",
+        values=("Debug", "Release"),
```
aidanheerdegen, line 28: No option for `RelWithDebInfo`?

packages/access-cice/package.py
anton-seaice: The CMake build will pick the correct "driver" code to work with dependencies built in access3-share

a.k.a the customisation in our fork are not essential for the model to build, but misses some nice to have customisation in our fork

I am not sure thats the question ?

packages/access3-share/package.py
```diff
@@ -0,0 +1,55 @@
...
+    )
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
+    depends_on("esmf@8.3.0:")
+    depends_on("esmf cflags='-fp-model precise' fflags='-fp-model precise'", when="%intel")
+    depends_on("fortranxml@4.1.2:")
+
+    depends_on("parallelio@2.5.10: build_type==RelWithDebInfo")
```
aidanheerdegen, line 38: `build_type` hard-coded.

packages/access-cice/package.py
```diff
@@ -0,0 +1,66 @@
...
+        default="NetCDF",
+        values=("NetCDF", "PIO", "Binary"),
+        description="CICE IO Method"
+    )
+
+    variant(
+        "build_type",
+        default="Release",
+        description="The build type to build",
+        values=("Debug", "Release"),
```
aidanheerdegen, line 39: No `RelWithDebInfo` option?

packages/access-mom6/package.py
```diff
@@ -0,0 +1,62 @@
...
+    variant("openmp", default=False, description="Enable OpenMP")
+    variant("mom_symmetric", default=True, description="Use symmetric memory in MOM6")
+    variant("access3", default=True, description="Building MOM6 library with Access3share")
+    variant("cesmcoupled", default=False, description="Enable parameters with cesm coupled")
+
+    depends_on("access3-share", when="+access3")
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
+    depends_on("fms@2021.03: build_type==RelWithDebInfo precision=64 +large_file ~gfs_phys ~quad_precision")
```
aidanheerdegen, line 43: Hard codes `build_type` and there is no `build_type` variant as there is in other packages.

packages/access-mom6/package.py
```diff
@@ -0,0 +1,62 @@
...
+        args = [
+            self.define_from_variant("OPENMP", "openmp"),
+            self.define_from_variant("ENABLE_MOM_SYMMETRIC", "mom_symmetric"),
+            self.define_from_variant("ACCESS3_MOM6", "access3"),
+            self.define_from_variant("CESMCOUPLED", "cesmcoupled"),
+        ]
+
+        args.append(self.define("CMAKE_C_COMPILER", self.spec["mpi"].mpicc))
+        args.append(self.define("CMAKE_CXX_COMPILER", self.spec["mpi"].mpicxx))
+        args.append(self.define("CMAKE_Fortran_COMPILER", self.spec["mpi"].mpifc))
```
aidanheerdegen, line 60: These should no longer be necessary.

packages/access3-exe/package.py
```diff
@@ -0,0 +1,72 @@
...
+        # make configurations a cmake argument
+        buildConf = ";".join(self.spec.variants["configurations"].value)
+        args = [
+            self.define("BuildConfigurations",buildConf)
+        ]
+
+        # we need this for cmake to find MPI_Fortran
+        args.append(self.define("CMAKE_C_COMPILER", self.spec["mpi"].mpicc))
+        args.append(self.define("CMAKE_CXX_COMPILER", self.spec["mpi"].mpicxx))
+        args.append(self.define("CMAKE_Fortran_COMPILER", self.spec["mpi"].mpifc))
```
aidanheerdegen, line 70: Should no longer be needed I believe.

packages/access3-exe/package.py
```diff
@@ -0,0 +1,72 @@
...
+        description="ACCESS-OM3 configurations to build",
+    )
+    
+    variant("openmp", default=False, description="Enable OpenMP")
+    
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+
+    for configurations in ["CICE6", "MOM6-CICE6", "MOM6-CICE6-WW3", "CICE6-WW3"]:
+        depends_on("access-cice+cesmcoupled+access3")
```
aidanheerdegen, line 51: Is this short-hand for:
```python
depends_on("access-cice")
depends_on("cesmcoupled")
depends_on("access3")
```

packages/access3-exe/package.py
aidanheerdegen, line 1: Is there a reason to have this separate package? Couldn't all this be just be put in the `access-mom6` package?

packages/access3-exe/package.py
```diff
@@ -0,0 +1,72 @@
...
+    submodules = True
+    maintainers = ["anton-seaice", "harshula", "micaeljtoliveira"]
+
+    license("Apache-2.0", checked_by="anton-seaice")
+
+    variant(
+        "build_type",
+        default="Release",
+        description="The build type to build",
+        values=("Debug", "Release"),
```
aidanheerdegen, line 25: No option for `RelWithDebInfo`?

Doesn't the `CMakePackage` provide this variant automatically anyway?

https://spack.readthedocs.io/en/latest/build_systems/cmakepackage.html#cmake-build-type

packages/access3-exe/package.py
aidanheerdegen: That is the extent of my high-level feedback. The rest is pretty low level I'm afraid.

aidanheerdegen (commented):
Hope that is helpful

packages/access3-exe/package.py
anton-seaice: Did you leave this comment on the correct file ?

`access3-exe` uses different code to `access-mom6` ...

We will at some point build models which don't include MOM6

packages/access3-exe/package.py
anton-seaice: Thanks - correct !

packages/access-cice/package.py
anton-seaice: I fear this turns off some optimisation which needs to be off for bfb repro, Ill investigate

packages/access3-exe/package.py
anton-seaice: No cesmcoupled and access3 are variants in the access-cice package

packages/access3-exe/package.py
aidanheerdegen: Sorry, I make a mistake in my comment:

> Is there a reason to have this separate package? Couldn't all this be just be put in the access-om3 package?

Does that make more sense?

anton-seaice: I don't understand how versioning and CD would work if we did this.

i.e. 

the version on the access-om3 deployment, is the version of a git tag in the access-nri/access-om3 release

the version on access3-exe is the version from the source code, i.e. the access-nri/access3-share repo

So I think we need a bundle package so we have something to version ? (ping @CodeGat )

harshula:
Hi @anton-seaice , Use the `maintainers()` function instead of the old way of defining a variable.

packages/access-mom6/package.py
harshula: Hi @anton-seaice , Any build problems after you removed this?

anton-seaice: It looks liek WW3 will still gives issues , similar to https://github.com/ACCESS-NRI/ACCESS-OM3/pull/52#issuecomment-2680394102, ill confirm tomorrow

packages/access-mom6/package.py
```diff
@@ -0,0 +1,49 @@
...
+    homepage = "https://github.com/ACCESS-NRI/MOM6"
+    git = "https://github.com/ACCESS-NRI/MOM6.git"
+    submodules = True
+    maintainers("minghangli-uni", "harshula")
+    
+    # see license file in https://github.com/ACCESS-NRI/MOM6/blob/e92c971084e185cfd3902f18072320b45d583a54/LICENSE.md
+    license("LGPL-3.0", checked_by="minghangli-uni")
+
+    variant("openmp", default=False, description="Enable OpenMP")
+    variant("mom_symmetric", default=True, description="Use symmetric memory in MOM6")
```
anton-seaice, line 25:
```suggestion
    variant("mom_symmetric", default=False, description="Use symmetric memory in MOM6")
```

harshula:
Hi @anton-seaice , Upstream have reduced the copyright header down to:
```
# Copyright Spack Project Developers. See COPYRIGHT file for details.
#
# SPDX-License-Identifier: (Apache-2.0 OR MIT)
```

packages/access3-exe/package.py
aidanheerdegen: > the version on the access-om3 deployment, is the version of a git tag in the access-nri/access-om3 release

Yep. Importantly we're talking about the deployment repo, not this spack package which has no versioning associated with it, except for the version of the entire `spack-packages` repo.

> the version on access3-exe is the version from the source code, i.e. the access-nri/access3-share repo

I don't think I understand this distinction. 

Maybe we need a third party to come and explain to each other what we're missing? @dougiesquire?

anton-seaice: > Importantly we're talking about the deployment repo, not this spack package which has no versioning associated with it

There is a sort of versioning for this spack package in the deployment. The access-om3 bundle package uses git(https://github.com/access-nri/access-om3) and then when a deployment is made it is given the version specified in the top-level spec of the environment file in that repo. E.g. [this](https://github.com/ACCESS-NRI/ACCESS-OM3/blob/e5f7419cb3a4edff5985fa64430f5778b332e7e7/spack.yaml#L9):

```
spack:
  specs:
    - access-om3@git.2025.01.1
```

when merged creates the [2025.01.1 release on github](https://github.com/ACCESS-NRI/ACCESS-OM3/releases/tag/2025.01.1)

It would work for spack to remove the bundle package, and make the top-level spec something like this:

```
spack:
  specs:
    - 'access3-exe@git.2025.03.0 configurations=MOM6,MOM6-CICE6,MOM6-CICE6-WW3'
```

However where would the version for the github release be set ? The release and it's tag would exist on the access-om3 repository, but there is no package using that repository without the bundle package

aidanheerdegen: @harshula and I just had an IRL chat, and we've come to the conclusion that we the top level package is for deployment and should be an bundle package, so pretty much ignore all my comments above, and I'll resolve this conversation. Thanks for engaging though, and sorry if that wasted your time.

packages/access3-exe/package.py
```diff
@@ -0,0 +1,68 @@
...
+    "MOM6",
+    "CICE6",
+    "WW3",
+    "MOM6-WW3",
+    "MOM6-CICE6",
+    "CICE6-WW3",
+    "MOM6-CICE6-WW3",
+)
+
+class Access3Exe(CMakePackage):
```
anton-seaice, line 18: @harshula suggested maybe a different name.

It could be `Access3Bin` or `Access3Model` ?

dougiesquire: `Access3`?

anton-seaice: In my view it's too similar to Access3Share ?

anton-seaice:
This is ready for final review:

See example deployments:
normal:https://github.com/ACCESS-NRI/ACCESS-OM3/pull/35#issuecomment-2738939605
with openmp: https://github.com/ACCESS-NRI/ACCESS-OM3/pull/67#issuecomment-2738951652


And standalone builds:
MOM standalone: https://github.com/ACCESS-NRI/ACCESS-OM3/pull/64#issuecomment-2738734621
CICE standalone: https://github.com/ACCESS-NRI/ACCESS-OM3/pull/65#issuecomment-2738776555
WW3 standalone: https://github.com/ACCESS-NRI/ACCESS-OM3/pull/66#issuecomment-2738776626


I can get this to produce the same answers as the old build system built with oneapi - see https://github.com/ACCESS-NRI/access-om3-configs/pull/164#issuecomment-2739033749 (although the bounds of this test are a little arbitrary, as it uses %oneapi compiler)

@aidanheerdegen @harshula @dougiesquire 

general FYI @ACCESS-NRI/ocean

packages/access-mom6/package.py
anton-seaice: There was some missing CMake for find_package(MPI) and linking that was needed for this to work. I have removed these lines

packages/access3-exe/package.py
anton-seaice: I've removed these lines :-)

packages/access-cice/package.py
anton-seaice: i removed this, it doesn't change answers

we can set this at the environment level if we want

packages/access3-share/package.py
anton-seaice: I removed this - as you noted. This is provided in the CMake super class

packages/access3-share/package.py
anton-seaice: I removed this,

packages/access-cice/package.py
anton-seaice: I removed this - as you noted - This is provided in the CMake super class

packages/access-mom6/package.py
anton-seaice: I removed this - as you noted - build_type is provided in the CMake super class

dougiesquire:
Thanks @anton-seaice, just a note that I have seen this but I'm going to hold off reviewing until the dependency PRs are resolved:
- https://github.com/ACCESS-NRI/MOM6/pull/2
- https://github.com/ACCESS-NRI/WW3/pull/2
- https://github.com/ACCESS-NRI/CICE/pull/6
- https://github.com/ACCESS-NRI/access3-share/pull/2

anton-seaice:
We tweaked some variable names - all the 5 variations i've been testing still build :)

packages/access3-exe/package.py
anton-seaice: Ok I changed it to just Access3

aidanheerdegen:
Maybe ping me again once everything is ready and @dougiesquire has a crack if that's ok. I don't have enough in-depth knowledge about the frankly dizzying array of builds you plan to be able to do, but I can have a squizz and see if I spot anything.

packages/access3/package.py
```diff
@@ -0,0 +1,68 @@
...
+    # force user to supply a build combination
+    conflicts(
+        "configurations=none",
+        msg=f"A configurations variant must be set, can be one or many of {KNOWN_CONF}"
+    )
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("access3-share")
+    depends_on("esmf@8.3.0:")
```
dougiesquire, line 49: `Access3Share` has `depends_on("esmf@8.7.0:")`
```suggestion
    depends_on("esmf@8.7.0:")
```

packages/access3/package.py
```diff
@@ -0,0 +1,68 @@
...
+        multi=True,
+        description="ACCESS-OM3 configurations to build",
+        sticky=True  # force concretizer to not pick alternative variants
+    )
+
+    # force user to supply a build combination
+    conflicts(
+        "configurations=none",
+        msg=f"A configurations variant must be set, can be one or many of {KNOWN_CONF}"
+    )
```
dougiesquire, line 44: I haven't tested any of this logic. Would that be helpful or are you confident this is doing what you want?

packages/access3/package.py
```diff
@@ -0,0 +1,68 @@
...
+    "MOM6-CICE6",
+    "CICE6-WW3",
+    "MOM6-CICE6-WW3",
+)
+
+class Access3(CMakePackage):
+    """Executable build for ACCESS version 3 climate models. The exectuable is 
+    defined in Community Mediator for Earth Prediction Systems (CMEPS). Currently 
+    implemented for ACCESS-OM3, and in the future may support ACCESS-CM3 and 
+    ACCESS-ESM3. This is a companion pacakge to Access3Share which builds the
```
dougiesquire, line 22:
```suggestion
    ACCESS-ESM3. This is a companion package to Access3Share which builds the 
```

packages/access3/package.py
```diff
@@ -0,0 +1,68 @@
...
+    submodules = True
+    maintainers("anton-seaice", "harshula", "micaeljtoliveira")
+    license("Apache-2.0", checked_by="anton-seaice")
+
+    variant(
+        "configurations",
+        values=(*KNOWN_CONF, 'none'),
+        default='none',
+        multi=True,
+        description="ACCESS-OM3 configurations to build",
```
dougiesquire, line 36:
```suggestion
        description=(
            "ACCESS-OM3 configurations to build. When a model component "
            "is not included in a configuration, that component is replaced by "
            "a CDEPS data component."
        )
```

packages/access3/package.py
```diff
@@ -0,0 +1,68 @@
...
+        if "WW3" in conf:
+            depends_on("access-ww3+access3", when=f"configurations={conf}")
+
+    def cmake_args(self):
+
+        # make configurations a cmake argument
+        buildConf = ";".join(self.spec.variants["configurations"].value)
+
+        args = [
+            self.define("BuildConfigurations", buildConf),
```
dougiesquire, line 65: Might be worth explicitly setting `ACCESS3_LIB_INSTALL` here for clarity (and in case the default is changed I guess...)?

```suggestion
            self.define("BuildConfigurations", buildConf),
            self.define("ACCESS3_LIB_INSTALL", False),
```

dougiesquire (commented):
Thanks @anton-seaice. This is looking really good. A few comments/questions/suggestions below.

You'll see I have questions about some of the hardcoded options, particularly those for `when=%intel`. In general, I'm not sure I understand the logic for when options (or min versions for that matter) should be set in Cmake vs the SPR vs the spack env.

packages/access-mom6/package.py
```diff
@@ -0,0 +1,44 @@
...
+    variant("asymmetric_mem", default=False, description="Use asymmetric memory in MOM6")
+    variant("access3", default=True, description="Building MOM6 library with Access3share")
+
+    depends_on("access3-share", when="+access3")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
+    depends_on("fms@2021.03: precision=64 +large_file ~gfs_phys ~quad_precision")
+    depends_on("fms build_type==RelWithDebInfo", when="%intel")
```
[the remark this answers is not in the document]
anton-seaice: Oh thanks - no we don't need this. Rel and RelWithDebInfo should give the same results I think.

anton-seaice, line 32:
```suggestion
```

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+
+    license("Apache-2.0", checked_by="anton-seaice")
+
+    variant("openmp", default=False, description="Enable OpenMP")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
+    depends_on("esmf@8.7.0:")
+    depends_on("esmf fflags='-fp-model precise'", when="%intel")
```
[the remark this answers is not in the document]
anton-seaice: Yes - just for consistency with old builds. I haven't test without this for %intel. (And won't because well use %oneapi)

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+    depends_on("fortranxml@4.1.2:")
+
+    depends_on("parallelio@2.5.10:")
+    depends_on(("parallelio build_type==RelWithDebInfo "
```
anton-seaice, line 31:
```suggestion
    depends_on(("parallelio "
```

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+                "fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model precise' "
+                "cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'"),
+                when="%intel")
```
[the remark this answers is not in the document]
anton-seaice: yes just for consistency with old builds.

gcc should compile. I don't know what values need setting here for %gcc to get determinism etc.

packages/access3/package.py
anton-seaice: I spent days on this so I am fairly confident :)

packages/access-mom6/package.py
```diff
@@ -0,0 +1,44 @@
...
+
+    variant("openmp", default=False, description="Enable OpenMP")
+    variant("asymmetric_mem", default=False, description="Use asymmetric memory in MOM6")
+    variant("access3", default=True, description="Building MOM6 library with Access3share")
+
+    depends_on("access3-share", when="+access3")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
```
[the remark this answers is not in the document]
anton-seaice: This is the first version use in the cosima build system by the looks

https://github.com/COSIMA/access-om3/commit/d7a2e95353c552b10ca4df999c285dac69852a45

packages/access-mom6/package.py
```diff
@@ -0,0 +1,44 @@
...
+    depends_on("fms@2021.03: precision=64 +large_file ~gfs_phys ~quad_precision")
```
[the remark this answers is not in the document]
anton-seaice: Indeed, this would fail (possibly incorrectly) before 2022.03 due to precision=64 :

https://github.com/spack/spack/blob/bf6ea7b047092ddc2df7cfc28ff0c0b764e68097/var/spack/repos/builtin/packages/fms/package.py#L92-L96

packages/access-mom6/package.py
```diff
@@ -0,0 +1,43 @@
...
```
anton-seaice, line 31:
```suggestion
    depends_on("fms@2023.02: precision=64 +large_file ~gfs_phys ~quad_precision")
```

packages/access-cice/package.py
```diff
@@ -0,0 +1,45 @@
...
+        default="NetCDF",
+        values=("NetCDF", "PIO", "Binary"),
+        description="CICE IO Method"
+    )
+
+    depends_on("access3-share", when="+access3") 
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:", when="io_type=NetCDF")
+    depends_on("parallelio@2.5.10:", when="io_type=PIO")
```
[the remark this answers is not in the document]
anton-seaice: It quite hard to follow the netcdf versioning in CMake because the it uses the custom FindNetCDF.cmake file and netcdf-c and netcdf-fortran use different versions.

I think/assume that the `find_package(NetCDF 4.7.3)` in cmake, is referring the version of netcdf-c. And this line is referring to a version of netcdf-fortran. netcdf-fortran has a [dependency](https://github.com/spack/spack/blob/bf6ea7b047092ddc2df7cfc28ff0c0b764e68097/var/spack/repos/builtin/packages/netcdf-fortran/package.py#L42) on `netcdf-c@4.7.4:`

The versions should be the same or more restrictive here than in CMake so that the build fails earlier (i.e. in `spack concretize`) if there is an issue. 

All the versions here follow the cosima build, i do an adhoc test building with parallelio@2.5.3 which built without error, I didn't try running it.

packages/access-mom6/package.py
```diff
@@ -0,0 +1,43 @@
...
+    variant("openmp", default=False, description="Enable OpenMP")
+    variant("asymmetric_mem", default=False, description="Use asymmetric memory in MOM6")
+    variant("access3", default=True, description="Building MOM6 library with Access3share")
+
+    depends_on("access3-share", when="+access3")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
+    depends_on("fms@2023.02: precision=64 +large_file ~gfs_phys ~quad_precision")
```
anton-seaice, line 31: Did we decide we should use the access-fms package here ?

packages/access-cice/package.py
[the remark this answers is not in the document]
aidanheerdegen: Restrictions in the build system itself should only be the requirements for it to build (and run? run what?) successfully. 

Restrictions in the spack package should be related to the orchestration/packaging of builds. e.g. if there are known issues with a version of a dependency when `cice` is being built in a specific context (model). This shouldn't be that necessary, as the dependency version will flow down from the deployment environment (`spack.yaml`).

Opinions about specific versions for a model should be in the `spack.yaml`.

Is that clear enough? I guess some examples would be good.

This should be specified in the developer docs: https://github.com/ACCESS-NRI/spack-packages/issues/224

aidanheerdegen (commented):
I have pending comments, so had to submit a review for some reason.

packages/access3/package.py
```diff
@@ -0,0 +1,73 @@
...
+    variant(
+        "configurations",
+        values=(*KNOWN_CONF, 'none'),
+        default='none',
+        multi=True,
+        description=(
+            "ACCESS-OM3 configurations to build. When a model component "
+            "is not included in a configuration, that component is replaced by "
+            "a CDEPS data component."
+        )
```
anton-seaice, line 40:
```suggestion
        ),
```

packages/access-cice/package.py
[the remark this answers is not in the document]
anton-seaice: Sure - but I am not going to go and try building and running with netcdf-fortran@3.0.0 to find out if it works ... so I think it makes sense to set a minimum version ?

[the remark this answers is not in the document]
dougiesquire: > Restrictions in the spack package should be related to the orchestration/packaging of builds. e.g. if there are known issues with a version of a dependency when cice is being built in a specific context (model). This shouldn't be that necessary, as the dependency version will flow down from the deployment environment (spack.yaml).

@aidanheerdegen are you suggesting that the version restrictions in question here should be removed and the versions should be specified in the `spack.yaml` (they [are already](https://github.com/ACCESS-NRI/ACCESS-OM3/blob/5d3d92c7ee204d270f44ab471d838cd60617fc39/spack.yaml#L40-L49))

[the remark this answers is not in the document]
dougiesquire: > My suggestion is setting
> 
> `depends_on("parallelio@2.5.3:")` and leaving netcdf alone.
> 
> I can add a note to the spack if desired .

Fine with me, but I do want to hear what @aidanheerdegen thinks

packages/access-mom6/package.py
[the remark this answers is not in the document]
dougiesquire: That's pretty arbitrary. Shall we ditch the min version then? The exact version will be specified in the spack.yaml for model deployments

packages/access-ww3/package.py
```diff
@@ -0,0 +1,31 @@
...
+    maintainers("anton-seaice", "harshula")
+    license("LGPL-3.0", checked_by="anton-seaice")
+
+    variant("openmp", default=False, description="Enable OpenMP")
+    variant("access3", default=True, description="Install CICE as library for Access3 models") 
+
+    depends_on("access3-share", when="+access3")
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
```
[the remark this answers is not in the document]
dougiesquire: As for `access-mom6`, shall we ditch the min version? The exact version will be specified in the spack.yaml for model deployments

packages/access3-share/package.py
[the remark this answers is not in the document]
dougiesquire: Could we add a comment to that effect

packages/access3-share/package.py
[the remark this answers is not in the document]
dougiesquire: > yes just for consistency with old builds.

Could we add a comment to that effect

> gcc should compile. I don't know what values need setting here for %gcc to get determinism etc.

I was unable to compile `access-mom6` standalone with gcc. We should probably be explicit somewhere that we currently only test with intel compilers (maybe we already are?)

packages/access-mom6/package.py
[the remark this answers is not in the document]
dougiesquire: Sorry, just saw you've done this below

packages/access-mom6/package.py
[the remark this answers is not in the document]
anton-seaice: No we wan't stuff like build-ci to match against a newish version of netcdf

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+
+    license("Apache-2.0", checked_by="anton-seaice")
+
+    variant("openmp", default=False, description="Enable OpenMP")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
+    depends_on("esmf@8.7.0:")
+    depends_on("esmf fflags='-fp-model precise'", when="%intel")
```
anton-seaice, line 27:
```suggestion
    depends_on("esmf fflags='-fp-model precise'", when="%intel") # for consistency with access-om3-nuopc builds, e.g. https://github.com/ACCESS-NRI/spack-packages/blob/e2bdb46e56af8ac14183e7ed25da9235486c973a/packages/access-om3-nuopc/package.py#L58
```

packages/access-cice/package.py
anton-seaice, line 34:
```suggestion
    depends_on("parallelio@2.5.3:", when="io_type=PIO")
```

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+    variant("openmp", default=False, description="Enable OpenMP")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
+    depends_on("esmf@8.7.0:")
+    depends_on("esmf fflags='-fp-model precise'", when="%intel") # for consistency with access-om3-nuopc builds, e.g. https://github.com/ACCESS-NRI/spack-packages/blob/e2bdb46e56af8ac14183e7ed25da9235486c973a/packages/access-om3-nuopc/package.py#L58
+    depends_on("fortranxml@4.1.2:")
+
+    depends_on("parallelio@2.5.10:")
```
anton-seaice, line 30:
```suggestion
    depends_on("parallelio@2.5.3:")
```

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+    depends_on(("parallelio "
+                "fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model precise' "
+                "cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'"),
+                when="%intel")
```
anton-seaice, line 34:
```suggestion
    depends_on(("parallelio "
                "fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model precise' "
                "cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'"),
                when="%intel") # consistency with access-om3-nuopc builds, e.g. https://github.com/ACCESS-NRI/spack-packages/blob/e2bdb46e56af8ac14183e7ed25da9235486c973a/packages/access-om3-nuopc/package.py#L65
```

packages/access-mom6/package.py
dougiesquire: Nope, we want to stick with what you have

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+
+    license("Apache-2.0", checked_by="anton-seaice")
+
+    variant("openmp", default=False, description="Enable OpenMP")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
+    depends_on("esmf@8.7.0:")
+    depends_on("esmf fflags='-fp-model precise'", when="%intel") # for consistency with access-om3-nuopc builds, e.g. https://github.com/ACCESS-NRI/spack-packages/blob/e2bdb46e56af8ac14183e7ed25da9235486c973a/packages/access-om3-nuopc/package.py#L58
```
dougiesquire, line 27:
```suggestion
    depends_on("esmf fflags='-fp-model precise'", when="%intel")  # for consistency with access-om3-nuopc builds, e.g. https://github.com/ACCESS-NRI/spack-packages/blob/e2bdb46e56af8ac14183e7ed25da9235486c973a/packages/access-om3-nuopc/package.py#L58
```

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+    depends_on("fortranxml@4.1.2:")
+
+    depends_on("parallelio@2.5.3:")
+    depends_on(("parallelio "
+                "fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model precise' "
+                "cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'"),
+                when="%intel") # consistency with access-om3-nuopc builds, e.g. https://github.com/ACCESS-NRI/spack-packages/blob/e2bdb46e56af8ac14183e7ed25da9235486c973a/packages/access-om3-nuopc/package.py#L65
```
dougiesquire, line 34:
```suggestion
                when="%intel")  # consistency with access-om3-nuopc builds, e.g. https://github.com/ACCESS-NRI/spack-packages/blob/e2bdb46e56af8ac14183e7ed25da9235486c973a/packages/access-om3-nuopc/package.py#L65
```

dougiesquire (commented):
Fixes for failing linting checks

dougiesquire (dismissed):
This looks good from my perspective, thanks @anton-seaice. My preference is to wait for approval from @aidanheerdegen before merging, particularly in regards to the setting of versions.

anton-seaice:
Last call for reviews @aidanheerdegen and @harshula  :)

packages/access-mom6/package.py
```diff
@@ -0,0 +1,43 @@
...
+    depends_on("fms +openmp", when="+openmp")
+    depends_on("fms ~openmp", when="~openmp")
```
harshula, line 33: What's your intention with these three lines?

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+
+    license("Apache-2.0", checked_by="anton-seaice")
+
+    variant("openmp", default=False, description="Enable OpenMP")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
+    depends_on("esmf@8.7.0:")
+    depends_on("esmf fflags='-fp-model precise'", when="%intel")  # for consistency with access-om3-nuopc builds, e.g. https://github.com/ACCESS-NRI/spack-packages/blob/e2bdb46e56af8ac14183e7ed25da9235486c973a/packages/access-om3-nuopc/package.py#L58
```
harshula, line 27: Has the behaviour been verified?

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+    depends_on("fortranxml@4.1.2:")
+
+    depends_on("parallelio@2.5.3:")
+    depends_on(("parallelio "
+                "fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model precise' "
+                "cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'"),
+                when="%intel")  # consistency with access-om3-nuopc builds, e.g. https://github.com/ACCESS-NRI/spack-packages/blob/e2bdb46e56af8ac14183e7ed25da9235486c973a/packages/access-om3-nuopc/package.py#L65
```
harshula, line 34: Has the behaviour been verified?

packages/access3-share/package.py
anton-seaice: No - we're using oneapi

packages/access-ww3/package.py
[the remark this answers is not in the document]
anton-seaice: No we wan't stuff like build-ci to match against a newish version of netcdf, and for it to match against a netcdf version which meets the minimum set in CMakeLists

packages/access3-share/package.py
```diff
@@ -0,0 +1,42 @@
...
+    submodules = True
+    maintainers("anton-seaice", "harshula", "micaeljtoliveira")
+
+    license("Apache-2.0", checked_by="anton-seaice")
+
+    variant("openmp", default=False, description="Enable OpenMP")
+
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:")
```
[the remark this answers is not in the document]
anton-seaice: No we wan't stuff like build-ci to match against a newish version of netcdf, and for it to match against a netcdf version which meets the minimum set in CMakeLists

packages/access3-share/package.py
anton-seaice: No - we're using oneapi

packages/access3/package.py
```diff
@@ -0,0 +1,73 @@
...
+    depends_on("access3-share")
+    depends_on("esmf@8.7.0:")
+
+    for conf in KNOWN_CONF:
+        if "CICE6" in conf:
+            depends_on("access-cice+access3", when=f"configurations={conf}")
+        if "MOM6" in conf:
+            depends_on("access-mom6+access3", when=f"configurations={conf}")
+        if "WW3" in conf:
+            depends_on("access-ww3+access3", when=f"configurations={conf}")
```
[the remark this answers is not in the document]
anton-seaice: I promise this works :)

I think the issue you might be thinking of is trying to test against items in the spec, e.g. `if ( "configurations=CICE6" in spec.variants) ` which doesn't get picked up at the concretise step

packages/access-mom6/package.py
anton-seaice: This is from the `access-om3-nuopc` package.

These look like sensible defaults that well never change

Is there a more explicit question about the fms variants?

packages/access3/package.py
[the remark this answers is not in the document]
harshula: Peter's feedback from last night:
```
That should work yes: although it looks like you are allowing any of cice6, mom6, and ww3 to be enabled or disabled, which I think you could also achieve with boolean variants:

variant(cice6, default=True)
variant(mom6, default=True)
variant(ww3, default=False)
conflicts(~cice6~mom6~ww3) # IIRC you wanted at least one of these to be used

We've spoken about variant combinations before, so apologies if this is a repeat suggestion: did something come up for you when you tried boolean variants?
```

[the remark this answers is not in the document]
anton-seaice: We typically build "MOM6", "MOM6-CICE6" and "MOM6-CICE6-WW3" and not the other combinations.

So with boolean variants it would build all the combinations, but mostly we are only interested in those three

anton-seaice:
> * Docstring line length should be reduced. i.e. https://peps.python.org/pep-0008/#maximum-line-length
> 
> * Should the variant `access3` in all the SPRs have a consistently worded descriptions?

I made some changes for these - let me know if I missed anything

packages/access3/package.py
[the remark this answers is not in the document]
scheibelp: Thanks, I recall speaking about this in-depth, but didn't recall the full context.

Compared to boolean variants, this definitely makes it easier to control supported combos by changing `KNOWN_CONF`. The only potential issue I see is if other packages that depend on this package want /need to change their behavior based on which of these configurations is enabled.

With the boolean variants, one could say e.g.

```
conflicts(+cice6~mom6~ww3) # cice6 shouldn't be enabled by itself
```

although there's only an advantage if dependents want to check on whether these configs are enabled.

packages/access-mom6/package.py
harshula: From Peter:
> they stack in the sense that if we call this package foo, and you ask for foo+openmp, then it will depend on fms@2023.02: +openmp precision=64 +large_file ~gfs_phys ~quad_precision

harshula (approved):
Well done!
Please remember to squash the commits.

packages/access-cice/package.py
```diff
@@ -0,0 +1,50 @@
...
+        default="NetCDF",
+        values=("NetCDF", "PIO", "Binary"),
+        description="CICE IO Method"
+    )
+
+    depends_on("access3-share", when="+access3") 
+    depends_on("cmake@3.18:", type="build")
+    depends_on("mpi")
+    depends_on("netcdf-fortran@4.6.0:", when="io_type=NetCDF")
+    depends_on("parallelio@2.5.3:", when="io_type=PIO")
```
aidanheerdegen, line 39: Do you want to specify a minimum `netcdf-fortran` version when building with `parallelio`?

I noticed the [outstanding issue with `parallelio@2.6.2`](https://github.com/spack/spack/issues/42716) [still hasn't been resolved](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/parallelio/package.py#L62)..

Maybe the solution then is to make a PR to add that the source package?

aidanheerdegen (approved):
One question, otherwise happy for you to merge. 

One point for a possible future change is to keep all the common `depends_on` logic in a single place and include them into the packages. I guess a mixin class?

packages/access-cice/package.py
aidanheerdegen: Above isn't crucial, as the version will be set in the model deployment environment, but just want to point this out.

anton-seaice: We are currently using netcdf 4.9.2 in the access-om3 environment, so this doesn't impact OM3 directly. There is a solid workaround to that issue in manually specifying a newer netcdf version when requesting the new parallelio version.

dougiesquire (approved):
I resolved some review comments that I started, but there are still a few others started by others that I didn't feel qualified to resolve

PR #172, opened 2024-11-28
Migrate ISSM to GitHub repo

### LINKED ISSUE #167: ISSM: transition to GitHub repo
ISSM code was under subversion but is now hosted on GitHub

https://github.com/ISSMteam/ISSM

The spack package should be updated to use this as the code source

https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/issm/package.py

aidanheerdegen:
Add support for new [ISSM GitHub repo](https://github.com/ISSMteam/ISSM).

Closes #167

### DIFF
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -27,9 +27,7 @@ class Issm(AutotoolsPackage):
     depends_on("m1qn3")
 
     def url_for_version(self, version):
-        url = "https://github.com/ISSMteam/ISSM/archive/refs/tags/v{0}.tar.gz".format(version)
-        return url
-
+        return "https://github.com/ISSMteam/ISSM/tarball/v{0}".format(version)
 
     def autoreconf(self, spec, prefix):
         autoreconf("--install", "--verbose", "--force")
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -12,9 +12,10 @@ class Issm(AutotoolsPackage):
     """Ice-sheet and Sea-Level System Model"""
 
     homepage = "https://issm.jpl.nasa.gov/"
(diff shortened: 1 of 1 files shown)


aidanheerdegen:
As noted in the issue, compiles successfully with this change

https://github.com/ACCESS-NRI/spack-packages/issues/167#issuecomment-2505240221

But does require `gcc/13`: `spack install issm %gcc@13

Should this also be in the package recipe until https://github.com/Reference-ScaLAPACK/scalapack/issues/97 is resolved?

packages/issm/package.py
```diff
@@ -25,6 +26,11 @@ class Issm(AutotoolsPackage):
     depends_on("petsc+metis+mumps+scalapack")
     depends_on("m1qn3")
 
+    def url_for_version(self, version):
```
harshula, line 29: Does our usual syntax and github URL not work? i.e.
```
        return "https://github.com/ISSMteam/ISSM/tarball/{0}".format(version)
```

packages/issm/package.py
```diff
@@ -25,6 +26,11 @@ class Issm(AutotoolsPackage):
...
+        url = "https://github.com/ISSMteam/ISSM/archive/refs/tags/v{0}.tar.gz".format(version)
+        return url
```
harshula, line 31: Does our usual syntax and github URL not work? i.e.
```
        return "https://github.com/ISSMteam/ISSM/tarball/{0}".format(version)
```

aidanheerdegen: I don't think I tried that one. I did copy one from another package that was different again, and it definitely didn't work.

I have changed to the suggested format above and it works. Note that they're using `vX.X` as their tags, so I've made the change as above, but pre-pending a `v`.

harshula (approved):

PR #203, opened 2025-02-25
set fp-model precise for oneapi in om3-nuopc package

### LINKED ISSUE #202: %oneapi in access-om3-nuopc package
To build access-om3-nuopc package using oneapi, we need to set `fp-model precise` for the esmf package similar to 

https://github.com/ACCESS-NRI/spack-packages/blob/3bfca290c24b48c410503b1bf8228f4cabd6e857/packages/access-om3-nuopc/package.py#L58

I had a quick look at other packages and don't see any other changes needed. Ill make a PR

anton-seaice:
closes #202

### DIFF
--- a/packages/access-om3-nuopc/package.py
+++ b/packages/access-om3-nuopc/package.py
@@ -1,4 +1,4 @@
-# Copyright 2013-2022 Lawrence Livermore National Security, LLC and other
+# Copyright 2013-2025 Lawrence Livermore National Security, LLC and other
 # Spack Project Developers. See the top-level COPYRIGHT file for details.
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
@@ -15,12 +15,6 @@ class AccessOm3Nuopc(CMakePackage):
 
     version("main", branch="main", submodules=True)
 
-    variant(
-        "build_type",
-        default="Release",
-        description="The build type to build",
-        values=("Debug", "Release"),
-    )
     variant(
         "configurations",
         default="MOM6-CICE6, CICE6-WW3, MOM6-CICE6-WW3",
@@ -55,14 +49,18 @@ class AccessOm3Nuopc(CMakePackage):
     depends_on("mpi")
     depends_on("netcdf-fortran@4.6.0:")
     depends_on("esmf@8.3.0:")
-    depends_on("esmf cflags='-fp-model precise' fflags='-fp-model precise'", when="%intel")
+    depends_on("esmf fflags='-fp-model precise'", when="%intel")
     depends_on("fortranxml@4.1.2:")
-    depends_on("fms@2021.03: build_type==RelWithDebInfo precision=64 +large_file ~gfs_phys ~quad_precision")
+    depends_on("fms@2021.03: precision=64 +large_file ~gfs_phys ~quad_precision")
+    depends_on("fms build_type==RelWithDebInfo", when="%intel")
     depends_on("fms +openmp", when="+openmp")
     depends_on("fms ~openmp", when="~openmp")
 
-    depends_on("parallelio@2.5.10: build_type==RelWithDebInfo")
-    depends_on("parallelio fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model source' cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'", when="%intel")
+    depends_on("parallelio@2.5.10:")
+    depends_on(("parallelio build_type==RelWithDebInfo "
+                "fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model precise' "
+                "cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'"),
+                when="%intel")
 
     flag_handler = CMakePackage.build_system_flags
 


anton-seaice:
I missed the parallelio flags:

`depends_on("parallelio fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model source' cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'", when="%intel")`

about half these flags are not in the [oneapi programme guide](https://www.intel.com/content/www/us/en/docs/oneapi/programming-guide/2024-2/overview.html)

Are you or @manodeep familiar with their replacements ? Otherwise I will go digging :-)

manodeep:
@anton-seaice Flags that are not valid in oneAPI will be silently ignored - so, while not best practice, no harm done. There are two things I can see:

- The value of `fp-model` is different between `fflags` and `cflags` - should be set to `precise` for both
- the oneAPI compiler id for is `oneapi` under `spack` - might need to change that value.

manodeep:
I took. a quick look - the unsupported fortran flags seem to be: `-convert big_endian -assume byterecl -assume realloc_lhs`.  The alphabetical list of supported compiler options in oneAPI 2025 is here: https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2025-0/alphabetical-option-list.html

The `-assume realloc_lhs` option is no longer needed - [source](https://community.intel.com/t5/Intel-Fortran-Compiler/Describe-assume-realloc-lhs/td-p/1255621)

anton-seaice:
Thanks @manodeep 

`-convert big_endian` is kind of strange.  It sets the endianess for unformatted IO fortran IO. It looks like its only related to some GPTL timing stuff in the [parallelio library](https://github.com/NCAR/ParallelIO/blob/b38e34eeb9b75ce81ac94daf7c5245931de00b9d/tests/performance/kt.PIO1.perfmakefile#L13). Maybe its ok to turn that off? There doesn't seem to be a oneapi alternative, but there is a Runtime Environment variable of `FORT_CONVERSION = big_endian` which might do the same thing.
`-assume byterecl` seems to be similar - it sets the block size for unformatted IO to interpret `recl` sizes as bytes. Again - the places in the code using unformatted IO seem [to only be in tests](https://github.com/search?q=repo%3ANCAR%2FParallelIO+open%28&type=code&p=1), so I think its ok to change this. 

They seem to go back to this commit: https://github.com/COSIMA/spack-config/commit/3a1aca78cfeee62cdcffe2b174cb63b910c816f7

Any thoughts @micaeljtoliveira ?

micaeljtoliveira:
@anton-seaice All those flags come from the corresponding CESM builds and, IIRC, I was not able to get a working model that gave bit-wise identical results to the equivalent CESM model without them.

manodeep:
SInce the flags are not supported by the compiler directly, and were originally put in place for bitwise-identical outputs - would a reasonable strategy be to first try without the flags and see whether that breaks bitwise-identical outputs?

(My bias: I am not a fan of environment variables that might break default user expectations)

anton-seaice:
Yeah - ill give it at a go. We broke this in a different but as yet undiscovered way recently, so it might take a while !

packages/access-om3-nuopc/package.py
```diff
@@ -58,7 +58,7 @@ class AccessOm3Nuopc(CMakePackage):
 
     depends_on("parallelio@2.5.10: build_type==RelWithDebInfo")
     depends_on("parallelio fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model precise' cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'", when="%intel")
-    depends_on("parallelio fflags='-fp-model precise' cflags='-fp-model precise', when='%oneapi')
+    depends_on("parallelio fflags='-fp-model precise' cflags='-fp-model precise'", when='%oneapi')
```
manodeep, line 61: @anton-seaice To be consistent and to reduce the chances of non-deterministic behaviour, you should add the `-qno-opt-dynamic-align` to both the flags for oneAPI as well (it is [turned off by default on oneAPI](https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2025-0/qopt-dynamic-align-qopt-dynamic-align.html) but best to be explicit in case Intel change their defaults in the future)

anton-seaice: Thanks :-)

anton-seaice:
This change:

- sets `-fp-model precise` consistently for intel and oneapi in parallelio and esmf dependencies.
- sets other parallelio flags for oneapi. Some of these flags set for intel no longer exist.

There is an example deployment:

https://github.com/ACCESS-NRI/ACCESS-OM3/pull/37#issuecomment-2716310497

Which has model determinism and restart reproducability (https://github.com/ACCESS-NRI/access-om3-configs/pull/196#issuecomment-2718122976)

This is ready for review @dougiesquire and @manodeep and/or @micaeljtoliveira

anton-seaice:
The difference between `-fp-model source` and `-fp-model precise` are [here](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2025-0/fp-model-fp.html)

dougiesquire (commented):
It would be great to get a recommendation from the transformation team on how to proceed with the `-fp-model` option. Are we happy with `-fp-model precise` for everything?

manodeep:
I am fine with `-fp-model=precise` for everything.

packages/access-om3-nuopc/package.py
```diff
@@ -56,14 +50,16 @@ class AccessOm3Nuopc(CMakePackage):
...
     depends_on("fortranxml@4.1.2:")
     depends_on("fms@2021.03: build_type==RelWithDebInfo precision=64 +large_file ~gfs_phys ~quad_precision")
     depends_on("fms +openmp", when="+openmp")
     depends_on("fms ~openmp", when="~openmp")
 
     depends_on("parallelio@2.5.10: build_type==RelWithDebInfo")
-    depends_on("parallelio fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model source' cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'", when="%intel")
-
+    depends_on("parallelio fflags='-qno-opt-dynamic-align -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model precise' cflags='-qno-opt-dynamic-align -fp-model precise -std=gnu99'", when="%intel")
+    depends_on("parallelio fflags='-qno-opt-dynamic-align -fp-model precise' cflags='-qno-opt-dynamic-align -fp-model precise'", when='%oneapi')
```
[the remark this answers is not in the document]
anton-seaice: > There's some relevant discussion above, but I can't see any resolution on `-convert big_endian` / `-assume byterec` or anything about `-ftz`

As it passed the restart repro / determinism tests I didn't investigate further.

-ftz is the [default now](https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2025-0/ftz-qftz.html) 

`-convert big_endian` / `-assume byterec` appeared to be related to unformatted IO, which only seemed to be used for doing timing tests on the ParallelIO library and not using it.

dougiesquire:
> Since the flags are not supported by the compiler directly, and were originally put in place for bitwise-identical outputs - would a reasonable strategy be to first try without the flags and see whether that breaks bitwise-identical outputs?

@manodeep, did you mean:
1. remove those flags with `%intel` and see if that changes answers relative to a previous run?
2. remove those flags with `%openapi` and see if two runs produce the same answers? (@anton-seaice [has verified](https://github.com/ACCESS-NRI/access-om3-configs/pull/196#issuecomment-2718122976) this)

manodeep:
> > Since the flags are not supported by the compiler directly, and were originally put in place for bitwise-identical outputs - would a reasonable strategy be to first try without the flags and see whether that breaks bitwise-identical outputs?
> 
> @manodeep, did you mean:
> 
>     1. remove those flags with `%intel` and see if that changes answers relative to a previous run?
> 
>     2. remove those flags with `%openapi` and see if two runs produce the same answers? (@anton-seaice [has verified](https://github.com/ACCESS-NRI/access-om3-configs/pull/196#issuecomment-2718122976) this)

@dougiesquire I wouldn't go down the `%intel` path anymore for verifying - `%intel` is just waiting for retirement :). For the `%oneapi` configs, yes, I was suggesting to remove those flags (`-convert bigendian`, `-assume byterecl` and check for bitwise identical outputs (with presumably `%intel` configs). If the outputs are identical - which @anton-seaice seems to have confirmed, then no need for those two flags. 

`-ftz` is a flag to treat tiny (intermediate + results) numbers as 0 - that flags seems to be used in a few places. Setting that flag improves performance at the cost of some accuracy. Happy with that being added.

harshula:
Hi All, Can the decision/recommendation please be documented in `dev-docs`?

anton-seaice:
> `-ftz` is a flag to treat tiny (intermediate + results) numbers as 0 - that flags seems to be used in a few places. Setting that flag improves performance at the cost of some accuracy. Happy with that being added.

From https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2025-0/ftz-qftz.html

> <h3>Default</h3> 
>    
> -ftz or /Qftz | Denormal results are flushed to zero.  Every optimization option O level, except O0, sets [Q]ftz.
> -- | --
> 
> 
> 

Its not well written but I think this mean its already set

dougiesquire:
> For the %oneapi configs, yes, I was suggesting to remove those flags (-convert bigendian, -assume byterecl and check for bitwise identical outputs (with presumably %intel configs). If the outputs are identical - which @anton-seaice seems to have confirmed, then no need for those two flags.

We _don't_ get bitwise identical output to `%intel` configs. Answers change with the move to `%oneapi` (dropping those flags at the same time). @anton-seaice has shown we get determinism and restart repro with `%oneapi`.

manodeep:
> Hi All, Can the decision/recommendation please be documented in `dev-docs`?

@harshula I have looked to document such flag discussions/decisions but it hasn't been obvious which section these comments should go into. Do you mind creating a section for "Compilers" with a sub-section on "Compiler Flags"?

harshula:
Hi @manodeep , Are you unable to create a new section?

manodeep:
@harshula I do not know the logic/context behind how the wiki pages are organised - that's why I would prefer for y'all to create the empty sections and I can then fill those out.

harshula:
Hi @manodeep , Try it and see if you can work it out. If there's a problem, we can fix it up.

micaeljtoliveira:
Looking at the actual changes, I was wondering what it actually means to say that OM3 requires, e.g., ESMF compiled with `cflags='-fp-model precise' fflags='-fp-model precise'`. Does it mean that ESMF needs to be compiled **exactly** with those flags, or that the flags need to include the listed flags? If it's the former (which is what I'm assuming), this is a bit problematic, because it means there's no way to build OM3 with ESMF compiled with some additional flags. Does someone know

micaeljtoliveira:
Regarding the actual choice of `fp-model`, do we really know enough at this point to set this in stone and force it across all builds of OM3? Wouldn't it be better to set it in the spack environments so that one can more easily change this if needed?

anton-seaice:
> Looking at the actual changes, I was wondering what it actually means to say that OM3 requires, e.g., ESMF compiled with `cflags='-fp-model precise' fflags='-fp-model precise'`. Does it mean that ESMF needs to be compiled **exactly** with those flags, or that the flags need to include the listed flags? If it's the former (which is what I'm assuming), this is a bit problematic, because it means there's no way to build OM3 with ESMF compiled with some additional flags. Does someone know what is the behaviour here?

It looks like the fortan flags get appended:

https://github.com/spack/spack/blob/f82de718cd46fe4ad6f179cff605f43dafbbcd1b/var/spack/repos/builtin/packages/esmf/package.py#L248

and in a `spack-build-02-build-out.txt` file: 

`ESMF_F90COMPILEOPTS: -fp-model precise -O -fPIC -debug minimal -assume realloc_lhs -m64 -mcmodel=small -pthread -threads -qopenmp`

and the c flags do [not get used](https://github.com/spack/spack/blob/f82de718cd46fe4ad6f179cff605f43dafbbcd1b/var/spack/repos/builtin/packages/esmf/package.py#L247).

(example output:

`ESMF_CCOMPILEOPTS: -std=c99 -O -DNDEBUG -fPIC -debug minimal -qopenmp`)

(example output in /scratch/tm70/***/tmp/spack-stage/spack-stage-esmf-git.v8.7.0_8.7.0-sipqpkuqig3xsk5yt57fsrapzqafl5ng)

> Regarding the actual choice of `fp-model`, do we really know enough at this point to set this in stone and force it across all builds of OM3? Wouldn't it be better to set it in the spack environments so that one can more easily change this if needed?

I am happy to go either way here - its not hard to modify the spack package later either.

micaeljtoliveira:
> It looks like the fortan flags get appended:

Thanks @anton-seaice that's useful to know. Nevertheless, that's not quite what I was trying to understand. I guess my question was rather what would happen if someone tried to install OM3 while setting extra flags to ESMF, like `-fp-model precise -g`.

anton-seaice:
> while setting extra flags to ESMF

oh setting extra / conflicting flags in the `spack.yaml` environment file ?

anton-seaice:
Additions to the environment file seem to get added:

e.g.

`esmf@git.v8.7.0=8.7.0%oneapi@2025.0.4 cflags='-fp-model precise' fflags='-fp-model strict -g -fp-model precise' ...`

I've no idea how that gets interpreted by the compiler

micaeljtoliveira:
> Additions to the environment file seem to get added:

Okay, that seems to answer my question: the spack concretiser does let you use a dependency compiled with some extra flags. On the other hand, it's a potential problem that we can add conflicting flags. 

Anyway, all good from my side regarding this PR.

anton-seaice:
> and the c flags do [not get used](https://github.com/spack/spack/blob/f82de718cd46fe4ad6f179cff605f43dafbbcd1b/var/spack/repos/builtin/packages/esmf/package.py#L247).
> 
> (example output:
> 
> `ESMF_CCOMPILEOPTS: -std=c99 -O -DNDEBUG -fPIC -debug minimal -qopenmp`)

As the `cflags` do not get used in the ESMF spack package, I suggest we removed them ?

This means that `-fp-model precise` is not set for any C or CXX code. I don't know the ESMF code well enough to know if that is important, but do know the current build is deterministic and reproducible for different length of restarts.

micaeljtoliveira:
> This means that -fp-model precise is not set for any C or CXX code. I don't know the ESMF code well enough to know if that is important, but do know the current build is deterministic and reproducible for different length of restarts.

From the tests I did a couple of years ago, there was no need to set these flags to have deterministic results. I think all the ESMF code is in C++ with Fortran interfaces, so the C++ flags are what you need to check.

anton-seaice:
Setting cflags has no impact, so the C++ flags are the ESMF defaults which do not set `-fp-model`

`ESMF_CXXCOMPILEOPTS: -std=c++11 -O -DNDEBUG -fPIC -debug minimal -m64 -mcmodel=small -pthread -qopenmp`

micaeljtoliveira:
Have you tried setting `cxxflags` instead of `cflags`? For example `cxxflags='-qno-opt-dynamic-align -fp-model precise'`.

anton-seaice:
No - it probably works. But why? The current build is deterministic / repro without it ?

micaeljtoliveira:
Ah, right. Sorry, I missed this in your previous comment. In that case, yes, please remove all flags passed to ESMF. I would say that, as a general rule, we should only set flags here that are strictly necessary.

anton-seaice:
I tried building without any specific build flags set for oneapi in this spack package, and the resulting build is [deterministic and has restart repro](https://github.com/ACCESS-NRI/access-om3-configs/pull/196#issuecomment-2738255569), so I think this change is not needed at all. (Noting the model build set the flags in question in CMake e.g. https://github.com/COSIMA/access-om3/blob/58e40cb4381b094be739f408329ae0eaaaa99882/CMakeLists.txt#L99)

anton-seaice:
Building without this change at all seems fine:

https://github.com/ACCESS-NRI/ACCESS-OM3/pull/68

and 

https://github.com/ACCESS-NRI/access-om3-configs/pull/196#issuecomment-2738516448

So I think i will close this. Thanks for the feedback - its given useful context for #171 

If we need build_type=RelWithDebInfo for dependencies we can set in the spack.yaml.

anton-seaice:
Closing as no-change, thanks for your help in understanding all this :)

PR #210, opened 2025-03-10
ACCESS-ESM1.5 & ACCESS-ESM1.6: remove versions from dependencies

harshula:
* The versions do not need to be hardcoded in the SPRs because we define the required versions 
  in the `spack.yaml` Spack environment file.
* Removing hdf5 because it is a dependency of netcdf-c, not oasis3-mct.

### DIFF
--- a/packages/cice4/package.py
+++ b/packages/cice4/package.py
@@ -20,10 +20,9 @@ class Cice4(MakefilePackage):
 
     version("access-esm1.5", branch="access-esm1.5")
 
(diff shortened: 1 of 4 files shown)


harshula:
### Testing

https://github.com/ACCESS-NRI/ACCESS-ESM1.5/pull/28#issuecomment-2711980798

blimlim (approved):
Thanks @harshula. I'm not on top of the details, however the results from running all look good

penguian (approved):
The changes all look reasonable.

Issue #195, opened 2025-02-18
triangle dependency in spack compilation of ISSM


justinh2002:
### Description:

There is an issue with `spack` compilation of `ISSM` associated with the `triangle` dependency in the ISSM package. The ISSM package applies three different patches to the original `triangle` hosted on the [CMU/Netlib website](https://www.cs.cmu.edu/~quake/triangle.html). However, the `spack` version of `triangle` takes the original `triangle` hosted on CMU/Netlib and installs it, and the mirrored version on [GitHub of `triangle`](https://github.com/libigl/triangle) is not identical to those two.

### Steps to Reproduce:

1. Attempt to install the ISSM package with the triangle dependency from mirrored Github version.
2. Observe that the ISSM patches applied to the triangle dependency are incompatible with the mirrored version on GitHub.

### Expected Behavior:

The triangle dependency in the Github mirror should be consistent with the original version hosted on the CMU website and should correctly apply the necessary patches.

### Actual Behavior:

The mirrored version on GitHub of `triangle` is not identical to the original version hosted on the CMU/Netlib website, causing issues with the patches applied by the ISSM package.

### Misc. context

ISSM package version: 4.24
triangle dependency source: CMU website and GitHub mirror

justinh2002:
Note further that the `triangle` package in ISSM unzips a tarball that is hosted on the [ISSM website](https://issm.ess.uci.edu/files/externalpackages/triangle.zip), whereas the [Github mirror](https://github.com/libigl/triangle) is based off the package hosted in Netlib in the [voronoi directory](https://www.netlib.org/voronoi/index.html). Both tarball sources are identical as of writing, but the Github mirror does introduce some changes to the original source files.

Specifically, its main change is that it converts the source code (`triangle.c`, `triangle.h`) in C into C++ compilable code. 

For more context about triangle, see extraneous details on the [CMU website](https://www.cs.cmu.edu/~quake/triangle.html).

justinh2002:
Strategy is to destroy existing fork of `triangle` in ACCESS (https://github.com/ACCESS-NRI/access-triangle), then create a new repository that unzips the Netlib tarball (https://www.netlib.org/voronoi/index.html), applies the ISSM patches in three separate commits, then use `access-triangle` as the dependency for spack `package.py` of ISSM.

justinh2002:
see new repo:

 https://github.com/ACCESS-NRI/issm-triangle

PR #211, opened 2025-03-12
Access-triangle SPR

justinh2002:
This PR addresses #195, commits here are for an access-triangle SPR.


packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
...
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+# ----------------------------------------------------------------------------
+#
+#     spack install access-triangle-git
+#
+# You can edit this file again by typing:
+#
+#     spack edit access-triangle-git
+#
+# ----------------------------------------------------------------------------
```
harshula, line 13: Is this relevant or correct?

packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
...
+
+from spack.package import *
+import glob
+import os
+
+
+class AccessTriangle(MakefilePackage):
+    """Example Spack package for the ISSM Triangle library."""
```
harshula, line 21: It's not an example. You may want to mention who created the triangle library originally.

packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2013-2024 Lawrence Livermore National Security, LLC and other
+# Spack Project Developers. See the top-level COPYRIGHT file for details.
```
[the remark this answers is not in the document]
justinh2002: Is this with ref. to line 2 only? Or should line 1 be removed?

packages/access-triangle/package.py
justinh2002: not relevant - will remove

packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
...
+    """Example Spack package for the ISSM Triangle library."""
+
+    homepage = "https://github.com/ACCESS-NRI/issm-triangle"
+    git = 'https://github.com/ACCESS-NRI/issm-triangle.git'
+
+    version('main')
+
+    # # If on some systems 'gmake' is truly required, keep it. Otherwise, 'make'
+    # # is often sufficient because Spack sets MAKE appropriately.
+    depends_on('gmake',  type='build')
```
harshula, line 30: Remove it, if the build succeeds without it.

packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
...
+    
+    def url_for_version(self, version):
+        return "https://github.com/ACCESS-NRI/issm-triangle/archive/refs/heads/{0}.tar.gz".format(version)
+
+    def edit(self, spec, prefix):
+        """
+        This stage is where you typically patch or customize the makefile.
+        If the package comes with a pre-written Makefile that needs minimal
+        changes, you can do them here. Below, we just copy in the 'configs'
+        so that the build can find them in `self.build_directory`.
```
harshula, line 41: The important step is copying the files needed to build triangle to the same directory level as the source code. Why mention `self.build_directory` when it's not referred to in code?

packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
...
+    def build(self, spec, prefix):
+        """
+        This is where we actually call `make shared`.
+        Using MakefilePackage, you *could* rely on build_targets,
+        but we'll be explicit here so you can see it clearly.
+        """
+        with working_dir(join_path(self.stage.source_path, "src")):
+            make('shared')
+            
+        #raise an exception
```
harshula, line 61: Should you delete this line?

packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
...
+        src = join_path(self.stage.source_path, "src")
+
+        # Create prefix directories
+        mkdirp(prefix.include)
+        mkdirp(prefix.lib)
+
+        with working_dir(src):
+            # Make sure we see what's actually there, for debugging:
+            ls_output = Executable('ls')('-l', '.', output=str, error=str)
+            print("Files in build directory:\n", ls_output)
```
harshula, line 76: Delete or comment out. Spack and/or Python should have a native function call for listing the content of a directory. Use that instead of executing `ls -l`.

packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
...
+        This is where we actually call `make shared`.
+        Using MakefilePackage, you *could* rely on build_targets,
+        but we'll be explicit here so you can see it clearly.
+        """
+        with working_dir(join_path(self.stage.source_path, "src")):
+            make('shared')
+            
+        #raise an exception
+
+    def install(self, spec, prefix):
```
harshula, line 63: I assume the package's Makefile(s) do not contain an `install` target?

packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
...
+
+        with working_dir(src):
+            # Make sure we see what's actually there, for debugging:
+            ls_output = Executable('ls')('-l', '.', output=str, error=str)
+            print("Files in build directory:\n", ls_output)
+            
+            install('triangle.h', prefix.include)
+
+            for libfile in glob.glob("libtriangle.*"):
+                install(libfile, prefix.lib)
```
harshula, line 81: Explicitly list the library files that you want to install, if that is possible.

packages/access-triangle/package.py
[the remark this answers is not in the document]
harshula: Not really useful after the initial development phase is finished. Don't use a variant for build debugging. A debug variant is good for runtime debugging.

packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
...
+import os
+
+
+class AccessTriangle(MakefilePackage):
+    """Example Spack package for the ISSM Triangle library."""
+
+    homepage = "https://github.com/ACCESS-NRI/issm-triangle"
+    git = 'https://github.com/ACCESS-NRI/issm-triangle.git'
+
+    version('main')
```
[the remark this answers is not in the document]
justinh2002: since this library has patches pre-applied such that it differs from the original v1.6, I would like to think that we can use versioning here..? There might be a version('develop') that we could follow on with.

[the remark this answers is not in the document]
aidanheerdegen: It is probably a good idea to tag the original version as 1.6 for clarity, but you don't have to make it an official version to install if that isn't useful.

I think there should be some sort of versioning so it is clear what code was used in a build. If you're still testing and might make more changes then there isn't a need to tag yet. However if you think it is pretty stable I'd be tempted to tag it and add that as a version.

You could choose to continue with the current versioning and tag it 1.6.1, or adopt a different scheme to make it clear this is not a version from the developer.

packages/access-triangle/package.py
justinh2002: removed mention of that

packages/access-triangle/package.py
```diff
@@ -0,0 +1,86 @@
...
+
+    # # If on some systems 'gmake' is truly required, keep it. Otherwise, 'make'
+    # # is often sufficient because Spack sets MAKE appropriately.
+    depends_on('gmake',  type='build')
+    depends_on('libx11', type='link')
```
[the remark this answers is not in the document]
justinh2002: done in new commit

packages/access-triangle/package.py
justinh2002: commented out

packages/access-triangle/package.py
justinh2002: done, build does succeed

packages/access-triangle/package.py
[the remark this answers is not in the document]
justinh2002: 1.6-access1?

packages/access-triangle/package.py
harshula: Hi @justinh2002 , Have you used `os.listdir` before? If you are not deleting these lines, then please use the native functions.

packages/access-triangle/package.py
```diff
@@ -0,0 +1,87 @@
...
+    """Triangle is a two-dimensional mesh generator and Delaunay
+    triangulator. Triangle generates exact Delaunay triangulations,
+    constrained Delaunay triangulations, conforming Delaunay
+    triangulations, Voronoi diagrams, and high-quality triangular
+    meshes."""
+
+    homepage = "https://github.com/ACCESS-NRI/issm-triangle"
+    git = 'https://github.com/ACCESS-NRI/issm-triangle.git'
+
+    version('1.6-access1')
```
harshula, line 21: Does that work? If the source code comes from the `main` branch of the git repository, I suspect you should have something like this:
`version("1.6-access1", branch="main")`

harshula:
Hi @justinh2002 , SPRs use double quotes (`"`) instead of single quotes (`'`) for strings.

packages/access-triangle/package.py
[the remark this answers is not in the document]
aidanheerdegen: Personally I'd just go with 1.6.1 or use a CalVer style like we do with the models, e.g. `2025.03.1`. The 1.6 is just to be clear what version the initial commit corresponds to and need only be a tag.

packages/access-triangle/package.py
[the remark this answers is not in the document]
harshula: Both

packages/access-triangle/package.py
justinh2002: there's no make install target

packages/access-triangle/package.py
justinh2002: Just copying the way ISSM did it in bash - I would prefer it this way just to follow exactly how ISSM compiles triangle.

harshula: It's best not to copy a bash script when writing build system code. Create an explicit list of files you want to install in `prefix.lib` and use it.

justinh2002: Done now.

packages/access-triangle/package.py
justinh2002: naming convention is now 1.6.1 with branch = main

justinh2002: all the strings have been replaced by double quotes

packages/access-triangle/package.py
justinh2002: Yup, fixed that now

justinh2002:
I think I've addressed all the comments now, feel free to check the last version of the commit.

packages/access-triangle/package.py
```diff
@@ -0,0 +1,85 @@
+# Copyright Spack Project Developers. See COPYRIGHT file for details.
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+
+from spack.package import *
+import glob
```
harshula, line 7: You can delete this line now that glob is not used.

packages/access-triangle/package.py
```diff
@@ -0,0 +1,85 @@
...
+import os
```
harshula, line 8: Can you comment-out this import too, since it's only for debugging?

harshula (approved):
Thanks!

aidanheerdegen (approved):
LGTM

PR #213, opened 2025-03-17
access-om2: change the version back to latest

harshula:
It's better to be consistent across all the SBRs and use `latest`.


aidanheerdegen (approved):
Happy to go with this

PR #215, opened 2025-03-20
Switch to using the ACCESS-TEST-model component

### LINKED ISSUE #191: Modify access-test to use stub source code
We'd like the access-test package to contain some stub source code to testing of source code modifications propagating into builds

https://github.com/ACCESS-NRI/ACCESS-TEST/issues/26

There is a PR to add source code to the model deployment repo

https://github.com/ACCESS-NRI/ACCESS-TEST/pull/27

and the package will need to reference(and build) that source code.

aidanheerdegen:
Closes #191 

Required for https://github.com/ACCESS-NRI/ACCESS-TEST/issues/32

### DIFF
--- a/packages/access-test/package.py
+++ b/packages/access-test/package.py
@@ -20,7 +20,7 @@ class AccessTest(BundlePackage):
 
     variant("mpi", default=True, description="MPI build")
 
-    depends_on("access-test-model+mpi", when="+mpi", type="run")
-    depends_on("access-test-model~mpi", when="~mpi", type="run")
+    depends_on("access-test-model-component+mpi", when="+mpi", type="run")
+    depends_on("access-test-model-component~mpi", when="~mpi", type="run")
 
     # There is no need for install() since there is no code.
--- a/packages/access-test/package.py
+++ b/packages/access-test/package.py
@@ -18,9 +18,9 @@ class AccessTest(BundlePackage):
 
     version("latest")
 
-    variant("deterministic", default=False, description="Deterministic build.")
+    variant("mpi", default=True, description="MPI build")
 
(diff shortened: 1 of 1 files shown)


packages/access-test/package.py
```diff
@@ -18,9 +18,9 @@ class AccessTest(BundlePackage):
 
     version("latest")
 
-    variant("deterministic", default=False, description="Deterministic build.")
+    variant("mpi", default=True, description="MPI build")
 
-    depends_on("oasis3-mct+deterministic", when="+deterministic", type="run")
-    depends_on("oasis3-mct~deterministic", when="~deterministic", type="run")
+    depends_on("access-test-model+mpi", when="+mpi", type="run")
```
CodeGat, line 23: So does  `access-test-model` mean model in the sense that `access-om3-nuopc` is the model of `access-om3` (the SBR)?

aidanheerdegen: Kinda. It's *a* model component. In this case we only have one component for simplicity. 

`access-om3-nuopc` is similar in that sense, but isn't a great one to compare against because it is an "uber" model that has all the components as sub-modules. That is an architecture that is currently being unwound.

CodeGat: In that case, would it be better to give it the name `access-test-component`? I think that might be where my confusion is coming from.

aidanheerdegen: Ah right. Sure.

aidanheerdegen: Fixed in #66eed7

CodeGat (approved):

CodeGat (approved):

PR #226, opened 2025-04-10
access-fms: add variants to create shared/dynamic libraries

harshula:


packages/access-fms/package.py
```diff
@@ -31,6 +31,9 @@ class AccessFms(CMakePackage):
         default=True,
         description="Enable compiler definition -DINTERNAL_FILE_NML.",
     )
+    # To build a dynamic library, both `pic` and `shared` are required:
```
aidanheerdegen, line 34: If both `pic` and `shared` are required shouldn't the `shared` variant enable the `pic` variant by default?

aidanheerdegen:
> No description provided.

:(

packages/access-fms/package.py
harshula: Happy to do that, but it will diverge from the built-in FMS SPR.

harshula: I'll use the following technique:
```
$ rg 'requires.*\+pic' var/*
var/spack/repos/builtin/packages/libogg/package.py
34:    requires("+pic", when="+shared")

var/spack/repos/builtin/packages/pcre/package.py
54:    requires("+pic", when="+shared build_system=autotools")

var/spack/repos/builtin/packages/freetype/package.py
69:    requires("+pic", when="+shared build_system=autotools")
```

Issue #228, opened 2025-04-14
remove `examples/` directories from ISSM SPR


justinh2002:
ISSM environment is using 5-6 times the `inode` data size compared to other environments, to remove the `examples/` directory in the SPR.

justinh2002:
#184

PR #229, opened 2025-04-14
remove examples directory from ISSM's SPR

justinh2002:
remove `examples/` from `self.stage.source_path` #228, just keep standard serialised installation and build.

### DIFF
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -113,8 +113,3 @@ class Issm(AutotoolsPackage):
         # Run the normal Autotools install logic
         make("install", parallel=False)
 
-        # Copy only the examples directory directly into the prefix directory
-        examples_src = join_path(self.stage.source_path, 'examples')
-        examples_dst = join_path(prefix, 'examples')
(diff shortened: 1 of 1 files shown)


packages/issm/package.py
```diff
@@ -105,16 +106,10 @@ def configure_args(self):
...
     
     def install(self, spec, prefix):
         # Run the normal Autotools install logic
         make("install", parallel=False)
 
-        # Copy only the examples directory directly into the prefix directory
-        examples_src = join_path(self.stage.source_path, 'examples')
-        examples_dst = join_path(prefix, 'examples')
-        install_tree(examples_src, examples_dst)
-
```
aidanheerdegen, line 120: If you want to retain this capability for installs other than the deployments we're doing centrally I'd be fine with making this a variant, e.g. `+examples` and have it default to off.

aidanheerdegen (dismissed):
Looks good as-is, but if you want to make this a variant feel free to do so and ping me for another review.

aidanheerdegen (approved):
LGTM

PR #239, opened 2025-05-14
access-om3 SPRs: use build_system_flags as the flag handler

harshula:
* The default flag_handler is inject_flags. This method is opaque and verifying whether the flags were applied is difficult.
* Setting the flag_handler to build_system_flags should result in the flags appearing in the package's build logs.

### DIFF
--- a/packages/access-cice/package.py
+++ b/packages/access-cice/package.py
@@ -40,6 +40,8 @@ class AccessCice(CMakePackage):
 
     root_cmakelists_dir = "configuration/scripts/cmake"
 
+    flag_handler = build_system_flags
+
     def cmake_args(self):
         args = [
             self.define_from_variant("CICE_OPENMP", "openmp"),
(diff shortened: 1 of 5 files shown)


harshula:
Spack documentation: https://spack.readthedocs.io/en/latest/packaging_guide.html#compiler-flags
I checked with @scheibelp , during the Spack upstream meeting, whether this was the best way to make the compiler flags inserted via the Spack environment (spack.yaml) more transparent for CMake builds.

dougiesquire (commented):
Looks sensible to me. Should this be set in all our SPRs?

micaeljtoliveira (approved):
I would definitely recommend setting this for all packages.

harshula:
Hi @dougiesquire & @micaeljtoliveira , For CMake and Autotools build systems, yes, we should change the flag_handler to `build_system_flags`. There's another branch for the next batch of SPRs: https://github.com/ACCESS-NRI/spack-packages/tree/change-default-flag-handler2.  However, for Makefile build systems we can't `use build_system_flags` and the options are trickier.

harshula:
### Testing

Verified by @anton-seaice : https://github.com/ACCESS-NRI/ACCESS-OM3/pull/86#issuecomment-2881801978

PR #240, opened 2025-05-15
Autotools & CMake SPRs: use build_system_flags as the flag handler

harshula:
* The default flag_handler is inject_flags. This method is opaque
   and verifying whether the flags were applied is difficult.
* Setting the flag_handler to build_system_flags should result in the
   flags appearing in the package's CMake build logs.

### DIFF
--- a/packages/access-fms/package.py
+++ b/packages/access-fms/package.py
@@ -40,6 +40,8 @@ class AccessFms(CMakePackage):
     depends_on("netcdf-fortran")
     depends_on("mpi")
 
+    flag_handler = build_system_flags
+
     def url_for_version(self, version):
         return "https://github.com/ACCESS-NRI/FMS/tarball/{0}".format(version)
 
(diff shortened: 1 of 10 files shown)


harshula:
### Testing

@anton-seaice verified that PR https://github.com/ACCESS-NRI/spack-packages/pull/239 which uses build_system_flags as the flag handler resulted in the flags appearing in the logs.

micaeljtoliveira (approved):

dougiesquire (approved):

aidanheerdegen:
Should this one also be changed to remove the explicit Class reference?

https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/access-om3-nuopc/package.py#L66

harshula:
Hi @aidanheerdegen , Do you want it changed?

aidanheerdegen:
> Hi @aidanheerdegen , Do you want it changed?

I only pointed it out because you'd made a similar change in `fortranxml`.

harshula:
Hi @aidanheerdegen , I'm assigned as a maintainer of `fortranxml`, whereas I'm not for `access-om3-nuopc`.

PR #241, opened 2025-05-15
Add access-generic-tracers as a dependency of access-mom6

dougiesquire:
This PR adds `access-generic-tracers` as a dependency of `access-mom6`. It also adds a variant to the `access-generic-tracers` SPR to allow switching between depending on `access-fms` or `fms`. This defaults to `access-fms` for backwards compatibility.

Depends on:
- https://github.com/ACCESS-NRI/MOM6/pull/10

### DIFF
--- a/packages/access-mom6/package.py
+++ b/packages/access-mom6/package.py
@@ -36,6 +36,7 @@ class AccessMom6(CMakePackage):
     depends_on("fms@2023.02: precision=64 +large_file ~gfs_phys ~quad_precision")
     depends_on("fms +openmp", when="+openmp")
     depends_on("fms ~openmp", when="~openmp")
+    depends_on("access-generic-tracers ~use_access_fms")
 
     flag_handler = build_system_flags
 
--- a/packages/access-generic-tracers/package.py
+++ b/packages/access-generic-tracers/package.py
@@ -21,8 +21,15 @@ class AccessGenericTracers(CMakePackage):
     # TODO: Needs to be changed once changes to build system enter master.
     version("development", branch="development", preferred=True)
 
+    variant(
+        "use_access_fms",
+        default=True,
+        description="If True, depend on access-fms, otherwise depend on fms"
+    )
+
     depends_on("access-mocsy")
-    depends_on("access-fms")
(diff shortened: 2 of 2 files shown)


packages/access-mom6/package.py
```diff
@@ -36,6 +36,7 @@ class AccessMom6(CMakePackage):
     depends_on("fms@2023.02: precision=64 +large_file ~gfs_phys ~quad_precision")
```
anton-seaice, line 36: Should there be a comment on why we use fms rather than access-fms ?

packages/access-mom6/package.py
```diff
@@ -36,6 +36,7 @@ class AccessMom6(CMakePackage):
...
     depends_on("fms +openmp", when="+openmp")
     depends_on("fms ~openmp", when="~openmp")
+    depends_on("access-generic-tracers ~use_access_fms")
```
anton-seaice, line 39: Should this only be possible when version > 2025.02.000 ?

dougiesquire: Shoot sorry, merged after @harshula's approval. Yes, strictly this should be 2025.02. I'll open another PR now

packages/access-mom6/package.py
dougiesquire: What did you have in mind?

anton-seaice: I don't know the reason :)

dougiesquire: access-fms really only exists for MOM5 at the moment. And for if we find we need to make ACCESS-specific changes to FMS. I can put that in a comment

dougiesquire: Done here: [63d2a32](https://github.com/ACCESS-NRI/spack-packages/pull/242/commits/63d2a324a6d64030337a5359f1004cdebd8a4736)

packages/access-mom6/package.py
dougiesquire: Done here: https://github.com/ACCESS-NRI/spack-packages/pull/242/commits/b6c5f5508853318cd2885cb0fca2aedf90d77ee6

PR #242, opened 2025-05-15
access-om3 main dependencies: add versions and specify min versions

dougiesquire:
This PR sets the minimum FMS version in the access-generic-tracers package to 2025.02. This to include the export of FindNetCDF.cmake in the FMS cmake install dir. Thanks to @anton-seaice for catching this [here](https://github.com/ACCESS-NRI/spack-packages/pull/241#discussion_r2090350755).

ADDED:
In addition, this PR 
- adds versions to the `access-cice`, `access-mom6`, `access-ww3`, `access3-share` and `access3` SPRs
- specifies min versions of `access-cice`, `access-mom6`, `access-ww3` in the `access3` SPR

`access-generic-tracers` and `access-mocsy` are still in development. Versions will be added in a subsequent PR.

### DIFF
--- a/packages/access-mom6/package.py
+++ b/packages/access-mom6/package.py
@@ -36,7 +36,7 @@ class AccessMom6(CMakePackage):
     depends_on("fms@2023.02: precision=64 +large_file ~gfs_phys ~quad_precision")
     depends_on("fms +openmp", when="+openmp")
     depends_on("fms ~openmp", when="~openmp")
-    depends_on("access-generic-tracers ~use_access_fms")
+    depends_on("access-generic-tracers ~use_access_fms", when="@2025.02.001:")
 
     flag_handler = build_system_flags
 
--- a/packages/access-mom6/package.py
+++ b/packages/access-mom6/package.py
@@ -21,6 +21,9 @@ class AccessMom6(CMakePackage):
     # see license file in https://github.com/ACCESS-NRI/MOM6/blob/e92c971084e185cfd3902f18072320b45d583a54/LICENSE.md
     license("LGPL-3.0-only", checked_by="minghangli-uni")
 
+    version("2025.02.001", commit="a5f4397")
+    version("2025.02.000", commit="e088c8b")
+
     variant("openmp", default=False, description="Enable OpenMP")
     variant("asymmetric_mem", default=False, description="Use asymmetric memory in MOM6")
     variant(
--- a/packages/access-mom6/package.py
+++ b/packages/access-mom6/package.py
@@ -37,7 +37,7 @@ class AccessMom6(CMakePackage):
     depends_on("mpi")
     depends_on("netcdf-fortran@4.6.0:")
     depends_on("fms@2023.02: precision=64 +large_file ~gfs_phys ~quad_precision")
-    depends_on("fms +openmp", when="+openmp")
+    depends_on("fms@2025.02: +openmp", when="+openmp")
     depends_on("fms ~openmp", when="~openmp")
     depends_on("access-generic-tracers ~use_access_fms", when="@2025.02.001:")
 
--- a/packages/access-cice/package.py
+++ b/packages/access-cice/package.py
@@ -19,6 +19,9 @@ class AccessCice(CMakePackage):
     # see license file at https://github.com/CICE-Consortium/CICE
     license("LicenseRef-CICE", checked_by="anton-seaice")
 
+    version("CICE6.6.0-3", commit="2c444bd")
+    version("CICE6.6.0-1", commit="964a445")
+
     variant("openmp", default=False, description="Enable OpenMP")
     variant(
         "access3",
--- a/packages/access-generic-tracers/package.py
+++ b/packages/access-generic-tracers/package.py
@@ -21,6 +21,8 @@ class AccessGenericTracers(CMakePackage):
     # TODO: Needs to be changed once changes to build system enter master.
     version("development", branch="development", preferred=True)
 
+    # NOTE: access-fms@mom5 should be used in OM2, ESM1.5 and ESM1.6 to preserve
+    # answers with previous releases
     variant(
         "use_access_fms",
         default=True,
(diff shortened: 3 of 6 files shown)


anton-seaice:
Oh this isn't what I was thinking off at all 😅

dougiesquire:
Oh yeah, I misread your comment haha. But this is still a valid change.

You actual comment is also important. I'll add the version limit too

packages/access-mom6/package.py
```diff
@@ -33,7 +33,7 @@ class AccessMom6(CMakePackage):
     depends_on("cmake@3.18:", type="build")
     depends_on("mpi")
     depends_on("netcdf-fortran@4.6.0:")
-    depends_on("fms@2023.02: precision=64 +large_file ~gfs_phys ~quad_precision")
+    depends_on("fms@2025.02: precision=64 +large_file ~gfs_phys ~quad_precision")
     depends_on("fms +openmp", when="+openmp")
     depends_on("fms ~openmp", when="~openmp")
     depends_on("access-generic-tracers ~use_access_fms")
```
anton-seaice, line 39:
```suggestion
    depends_on("access-generic-tracers ~use_access_fms"), when = "@2025.02.001:"
```

Incredibly moot

dougiesquire:
@harshula, apologies @anton-seaice submitted a review of https://github.com/ACCESS-NRI/spack-packages/pull/241 just as I merged it. This PR addresses his comment and some more things.

Do we also want to set the min version of `access-fms` in `access-generic-tracers`, i.e.:
```diff
- depends_on("access-fms", when="+use_access_fms")
+ depends_on("access-fms@mom5-2025.05.000:", when="+use_access_fms")
```

packages/access-mom6/package.py
```diff
@@ -33,10 +33,10 @@ class AccessMom6(CMakePackage):
     depends_on("cmake@3.18:", type="build")
     depends_on("mpi")
     depends_on("netcdf-fortran@4.6.0:")
-    depends_on("fms@2023.02: precision=64 +large_file ~gfs_phys ~quad_precision")
+    depends_on("fms@2025.02: precision=64 +large_file ~gfs_phys ~quad_precision")
```
anton-seaice, line 36: You don't need this change anymore

dougiesquire: Yup sorry didn't get to this before knocking off yesterday, but will soon

anton-seaice:
> Do we also want to set the min version of `access-fms` in `access-generic-tracers`, i.e.:

I think setting this is good, as long as the tags in https://github.com/ACCESS-NRI/GFDL-generic-tracers continue to use the `mom5-` prefix 

(Arguably we could set minimum versions for the whole package too (well all the Access3 packages) but haven't.)

dougiesquire:
> I think setting this is good, as long as the tags in https://github.com/ACCESS-NRI/GFDL-generic-tracers continue to use the `mom5-` prefix

Thinking about this more, I don't think we can guarantee this. The `mom5-*` tags are on the `mom5` branch. There's a chance we may want to build from the `main` branch in the future

packages/access-mom6/package.py
dougiesquire: Now done

anton-seaice (approved):
Thanks

Maybe write in the commit description the justifications for the min versions

dougiesquire:
It's there already for the fms version commit. I'll add something to your commit when I merge (pending approval from @harshula )

dougiesquire:
Hmm Spack doesn't like the min versions: https://github.com/ACCESS-NRI/ACCESS-OM3/actions/runs/15060292158/job/42334082123#step:11:340

Need to sort this out before merging

dougiesquire:
@anton-seaice this is ready for review.

Test deployment is here: https://github.com/ACCESS-NRI/ACCESS-OM3/pull/95

anton-seaice (commented):
The packaging guide recommends using git hashes.

https://spack.readthedocs.io/en/latest/packaging_guide.html#git

It says:

"This download method is trusted. It is the recommended way to securely download from a Git repository."

Its not really clear why they are trusted ... google implies that a commit hash is a checksum of the content of the repo and is checked for integrity as a built in feature of git. (e.g. https://graphite.dev/guides/git-hash ?)

packages/access-mom6/package.py
```diff
@@ -36,7 +39,7 @@ class AccessMom6(CMakePackage):
     depends_on("fms@2023.02: precision=64 +large_file ~gfs_phys ~quad_precision")
     depends_on("fms +openmp", when="+openmp")
```
anton-seaice, line 40:
```suggestion
    depends_on("fms +openmp", when="+openmp")
    depends_on("fms@2025.02:", when="+openmp~access3")
```

Sorry this is another hidden dep I forgot about - see https://github.com/ACCESS-NRI/ACCESS-OM3/pull/100#issuecomment-2893858219

dougiesquire: How come it's not needed with `+access3`?

anton-seaice: Apparently it is - https://github.com/ACCESS-NRI/ACCESS-OM3/pull/102#issuecomment-2896046821

It's related to the CMake exports again ...

I added OpenMP C Component to the FMS exports in https://github.com/NOAA-GFDL/FMS/pull/1663

A fixed suggestion:

```suggestion
    depends_on("fms@2025.02: +openmp", when="+openmp")
```

(Also if relying on new fms is problematic, we could patch the fms spack package ? The dependency is only related to cmake exports)

dougiesquire: I think it's fine to rely on new FMS. Its good to keep our dependencies up to date anyway.

The only annoying thing is that 2025.02 isn't in the FMS spack package yet, but I think [this](https://github.com/ACCESS-NRI/ACCESS-OM3/blob/0cfa1823f1d591bece55f8cb6903831b6b06c352/spack.yaml#L53) is a fine approach for now.

anton-seaice (approved):
Thanks @dougiesquire - i spot checked some hashes and it looks like it concretises fine too

dougiesquire:
Successful test deployment with these changes [here](https://github.com/ACCESS-NRI/ACCESS-OM3/pull/95#issuecomment-2896149793), so merging

PR #246, opened 2025-06-03
Add gtracers version to access-mocsy

dougiesquire:
The name of the `mom5` version/branch in the `access-mocsy` SPR is a little confusing since this version is used for all `access-generic-tracer` builds (i.e. with both MOM5 and MOM6). Following discussion on Zulip, I've created a [new `gtracers` branch on `ACCESS-NRI/mocsy`](https://github.com/ACCESS-NRI/mocsy/tree/gtracers) that is the same as the existing `mom5` branch. This PR adds a corresponding `gtracers` version to the SPR.

We should delete the `mom5` version/branch in a few months.


packages/access-mocsy/package.py
```diff
@@ -17,7 +17,9 @@ class AccessMocsy(MakefilePackage):
     maintainers("harshula")
 
     version("master", branch="master")
-    version("mom5", branch="mom5", preferred=True)
+    version("gtracers", branch="gtracers", preferred=True)
+    # TODO: Remove the 'mom5' version. The 'mom5' version/branch has been renamed to 'gtracers'.
```
harshula, line 21: Was it "renamed" or copied?

dougiesquire: Copied - text now modified thanks

harshula (approved):

PR #247, opened 2025-06-04
ISSM: make +ad variant self-contained (no PETSc, no Love, tape-alloc on)”

justinh2002:
This PR lets the `+ad` (Automatic Differentiation) flavour of `issm` build and run out-of-the-box:

- Removes hidden PETSc dependency in sea-level sub-modules (`Love`), which broke the compile with `fatal error: petscblaslapack.h: No such file or directory.`

- Enables `--enable-tape-alloc` (upstream-recommended for CoDiPack performance / stability).

- Supplies an explicit `BLAS/LAPACK` path so the AD variant links cleanly on clusters that don’t inject it.

- Adds an optional `-lgfortran` link flag for mixed C/Fortran toolchains that drop the runtime.

- No change for the default “classic” PETSc-backed build.

Testing:

- `spack install issm +ad +wrappers %gcc@13` now completes on Gadi (OpenMPI 4.1.5)

### DIFF
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -87,6 +87,8 @@ class Issm(AutotoolsPackage):
     depends_on("parmetis", when="+wrappers")
     depends_on("python@3.9.2:", when="+wrappers", type=("build", "run"))
     depends_on("py-numpy", when="+wrappers", type=("build", "run"))
+    
+    flag_handler = build_system_flags
 
     # GCC 14 breaks on several C++17 constructs used in ISSM
     conflicts("%gcc@14:", msg="ISSM cannot be built with GCC versions above 13")
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -68,7 +68,7 @@ class Issm(AutotoolsPackage):
     depends_on("mpi")
 
     # Linear-algebra stack - only for the *non-AD* flavour
-    depends_on("petsc", when="~ad")
+    depends_on("petsc+metis+mumps+scalapack", when="~ad")
     depends_on("parmetis")
     depends_on("metis")
     depends_on("mumps")
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -1,7 +1,7 @@
 # Copyright Spack Project Developers. See COPYRIGHT file for details.
 #
 # Copyright 2023 Angus Gibson
-# Justin Kin Jun Hew, 2025 – AD‑enabled flavour
+# Justin Kin Jun Hew, 2025 – Wrappers, Examples, Versioning, AD‑enabled flavour
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -64,7 +64,7 @@ class Issm(AutotoolsPackage):
     depends_on("libtool", type="build")
     depends_on("m4", type="build")
 
-    # Core runtime deps
+    # Core build + runtime deps
     depends_on("mpi")
 
     # Linear-algebra stack - only for the *non-AD* flavour
@@ -110,9 +110,6 @@ class Issm(AutotoolsPackage):
                 "CXXFLAGS",
                 f"-g -O3 -fPIC {self.compiler.cxx11_flag} -DCODI_ForcedInlines",
             )
-
-        env.set("CMAKE_C_COMPILER", self.spec["mpi"].mpicc)
-        env.set("CMAKE_CXX_COMPILER", self.spec["mpi"].mpicxx)
 
     # --------------------------------------------------------------------
     # Autoreconf hook
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -68,7 +68,7 @@ class Issm(AutotoolsPackage):
     depends_on("mpi")
 
     # Linear-algebra stack - only for the *non-AD* flavour
-    depends_on("petsc+metis+mumps+scalapack", when="~ad")
+    depends_on("petsc~examples+metis+mumps+scalapack", when="~ad")
     depends_on("parmetis")
     depends_on("metis")
     depends_on("mumps")
@@ -123,13 +123,17 @@ class Issm(AutotoolsPackage):
     # Configure phase - construct ./configure arguments
     # --------------------------------------------------------------------
     def configure_args(self):
-        args = [
+        args = []
+        if "+ad" in self.spec:
+            args.append("--enable-ad")
+
+        args.extend([
             "--enable-debugging",
             "--enable-development",
             "--enable-shared",
             "--without-kriging",
             "--without-Love",
-        ]
+        ])
 
         # Linear-algebra backend
         if "+ad" in self.spec:
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -69,6 +69,7 @@ class Issm(AutotoolsPackage):
 
     # Linear-algebra stack - only for the *non-AD* flavour
     depends_on("petsc", when="~ad")
+    depends_on("parmetis")
     depends_on("metis")
     depends_on("mumps")
     depends_on("scalapack")
@@ -84,7 +85,6 @@ class Issm(AutotoolsPackage):
 
     # Optional extras controlled by +wrappers
     depends_on("access-triangle", when="+wrappers")
-    depends_on("parmetis", when="+wrappers")
     depends_on("python@3.9.2:", when="+wrappers", type=("build", "run"))
     depends_on("py-numpy", when="+wrappers", type=("build", "run"))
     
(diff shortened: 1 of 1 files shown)


packages/issm/package.py
```diff
@@ -1,126 +1,195 @@
...
+# Justin Kin Jun Hew, 2025 – Wrappers, Examples, Versioning, AD‑enabled flavour
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import *
 
 
 class Issm(AutotoolsPackage):
-    """Ice-sheet and Sea-Level System Model"""
+    """Ice‑sheet and Sea‑Level System Model.
```
harshula, line 12: Why did you change the ASCII 0x2d character to Unicode 0x2011 character?

packages/issm/package.py
```diff
@@ -1,126 +1,195 @@
...
+    This recipe supports two distinct build flavours:
+
+    * **Classic** (default) – links against PETSc for linear algebra.
+    * **Automatic Differentiation** (+ad) – uses CoDiPack + MediPack and
+      **excludes PETSc** (ISSM’s AD implementation is not PETSc‑compatible).
+    """
 
     homepage = "https://issm.jpl.nasa.gov/"
-    git = "https://github.com/ACCESS-NRI/ISSM.git"
+    git      = "https://github.com/ACCESS-NRI/ISSM.git"
```
harshula, line 22: Please don't make these types of changes. Spack appears to standardise on the `black` linter. Perhaps, run `black` over your file and stick to that formatting.

packages/issm/package.py
harshula: Thanks for fixing that!

packages/issm/package.py
```diff
@@ -1,126 +1,181 @@
...
+
+    # Core runtime deps
     depends_on("mpi")
-    depends_on("petsc+metis+mumps+scalapack")
+
+    # Linear‑algebra stack – only for the *non‑AD* flavour
+    depends_on("petsc", when="~ad")
+    depends_on("metis")
+    depends_on("mumps")
+    depends_on("scalapack")
```
harshula, line 64: `depends_on("petsc+metis+mumps+scalapack")` is not the equivalent of the above. Can you please explain the motivation for this change?

packages/issm/package.py
```diff
@@ -1,126 +1,181 @@
...
+    # ──────────────────────────────────────────────────────────────────────
+    def setup_build_environment(self, env):
+        if "+ad" in self.spec:
+            # CoDiPack’s performance tips: force inlining & keep full symbols
+            env.append_flags(
+                "CXXFLAGS", f"-g -O3 -fPIC {self.compiler.cxx11_flag} -DCODI_ForcedInlines"
+            )
 
+        env.set("CMAKE_C_COMPILER", self.spec["mpi"].mpicc)
+        env.set("CMAKE_CXX_COMPILER", self.spec["mpi"].mpicxx)
```
harshula, line 104: Why do you need this? Spack has started to remove these because the relevant information can be set in the spack-config.

packages/issm/package.py
```diff
@@ -1,126 +1,181 @@
...
+
+    # ──────────────────────────────────────────────────────────────────────
+    # Build environment – inject AD compiler flags when needed
+    # ──────────────────────────────────────────────────────────────────────
+    def setup_build_environment(self, env):
+        if "+ad" in self.spec:
+            # CoDiPack’s performance tips: force inlining & keep full symbols
+            env.append_flags(
+                "CXXFLAGS", f"-g -O3 -fPIC {self.compiler.cxx11_flag} -DCODI_ForcedInlines"
+            )
```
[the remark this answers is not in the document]
justinh2002: apparently not - this is what's been asked of upstream to configure ISSM with ad

packages/issm/package.py
```diff
@@ -1,126 +1,181 @@
...
+    depends_on("codipack", when="+ad")
+    depends_on("medipack", when="+ad")
+
+    # Optional extras controlled by +wrappers
+    depends_on("access-triangle", when="+wrappers")
+    depends_on("parmetis", when="+wrappers")
     depends_on("python@3.9.2:", when="+wrappers", type=("build", "run"))
     depends_on("py-numpy", when="+wrappers", type=("build", "run"))
 
-    flag_handler = build_system_flags
```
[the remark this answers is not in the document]
justinh2002: it's not necessary

packages/issm/package.py
justinh2002: removed

packages/issm/package.py
```diff
@@ -1,126 +1,195 @@
...
+    # Dependencies
+    # ──────────────────────────────────────────────────────────────────────
+    # Build‑time tools
     depends_on("autoconf", type="build")
     depends_on("automake", type="build")
     depends_on("libtool",  type="build")
     depends_on("m4",       type="build")
+
+    # Core runtime deps
     depends_on("mpi")
```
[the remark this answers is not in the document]
justinh2002: yes it is a build + runtime dep

packages/issm/package.py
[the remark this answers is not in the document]
harshula: It was added for a good reason. Refer to the commit and PR that correspond with that line of code. Can you please add it back?

packages/issm/package.py
harshula: 1) Can you please read the documentation and make sure you understand what `depends_on("petsc+metis+mumps+scalapack")` does?
2) Why does ISSM need these libraries?

packages/issm/package.py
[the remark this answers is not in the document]
justinh2002: AD is incompatible with petsc, so the install is done a little differently, something like:

remove --with-petsc-dir,
add export CXXFLAGS="-g -O3 -fPIC -std=c++11 -DCODI_ForcedInlines",
add--with-codipack-dir="$ISSM_DIR/externalpackages/codipack/install" \--with-medipack-dir="$ISSM_DIR/externalpackages/medipack/install" \,

similar to the instructions here: https://issm.ess.uci.edu/trac/issm/wiki/totten#InstallingISSMwithCoDiPackAD

packages/issm/package.py
justinh2002: please see https://issm.ess.uci.edu/trac/issm/wiki/totten#InstallingISSMwithCoDiPackAD

packages/issm/package.py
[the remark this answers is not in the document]
harshula: Can you please explain why you can't do that in the ISSM source package's build system? Have you tried adding an `--enable-ad` option in ISSM's configure.ac that sets the relevant compiler flags?

packages/issm/package.py
justinh2002: AD doesn't work with `petsc` but it does require the other three libraries to build AD in the non-petsc build.

packages/issm/package.py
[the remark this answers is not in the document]
justinh2002: extensive changes would be required in `configure.ac`, it should work but why not just use spack's built-in `setup_run_environment`?

packages/issm/package.py
headmetal: I guess that works in the "~ad" case, but doesn't fix the "ad" case.

justinh2002: done

packages/issm/package.py
```diff
@@ -1,126 +1,193 @@
...
+    # --------------------------------------------------------------------
+    # Build environment - inject AD compiler flags when needed
+    # --------------------------------------------------------------------
+    def setup_build_environment(self, env):
+        if "+ad" in self.spec:
+            # CoDiPack's performance tips: force inlining & keep full symbols
+            env.append_flags(
+                "CXXFLAGS",
+                f"-g -O3 -fPIC {self.compiler.cxx11_flag} -DCODI_ForcedInlines",
+            )
```
harshula, line 114: Please add a comment with the URL of the document that recommended these flags.

justinh2002: done

packages/issm/package.py
```diff
@@ -1,126 +1,193 @@
...
+    # Automatic-differentiation libraries
+    depends_on("codipack", when="+ad")
+    depends_on("medipack", when="+ad")
+
+    # Optional extras controlled by +wrappers
+    depends_on("access-triangle", when="+wrappers")
     depends_on("python@3.9.2:", when="+wrappers", type=("build", "run"))
     depends_on("py-numpy", when="+wrappers", type=("build", "run"))
-
+
```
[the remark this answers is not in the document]
justinh2002: resolved

harshula (approved):
Thanks! Please remember to use squash merge.

PR #248, opened 2025-06-10
um7: add variant to support incremental builds

harshula:
@JhanSrbinovsky requested an option to do incremental builds. The documentation at https://metomi.github.io/fcm/doc/user_guide/build.html says `-f` option forces a "full (fresh) build". This PR provides a way to disable it.

### DIFF
--- a/packages/um7/package.py
+++ b/packages/um7/package.py
@@ -32,6 +32,12 @@ class Um7(Package):
     depends_on("netcdf-fortran@4.5.2:", type=("build", "link"))
     depends_on("oasis3-mct", type=("build", "link"))
 
+    # https://metomi.github.io/fcm/doc/user_guide/build.html
+    variant(
+        "full",
+        default=True,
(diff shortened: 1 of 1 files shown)


blimlim:
@harshula successful build with `~full` here: https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/90#issuecomment-2957890293

harshula:
FYI, `~full` is probably only relevant when using `spack develop`.

JhanSrbinovsky:
This seems to be adding functionality FOR a full build? From the 20 minute build time I thought this is what it was already doing by default?

JhanSrbinovsky:
@harshula My understanding was that you had realised this was limited by the indepence between releases, ddi you find a way around this?

harshula:
Hi @JhanSrbinovsky , I can't think of any workaround for completely independent builds to be forced to build incrementally. Only incremental builds like `spack develop` will benefit from this variant.

PR #262, opened 2025-07-01
access-mocsy: Add CMake build_system (default)

dougiesquire:
This PR adds support to the access-mocsy SPR for the CMake build system added in https://github.com/ACCESS-NRI/mocsy/pull/1.

Note, the `"master"` `version` is also removed, since neither the CMake nor existing Makefile build system support building the master branch.

### DIFF
--- a/packages/access-mocsy/package.py
+++ b/packages/access-mocsy/package.py
@@ -6,23 +6,53 @@
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
 from spack.package import find_libraries, install, join_path, mkdirp
+from spack.build_systems import cmake, makefile
 
-# https://spack.readthedocs.io/en/latest/build_systems/makefilepackage.html
-class AccessMocsy(MakefilePackage):
+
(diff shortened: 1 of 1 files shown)


JorgeG94:
lgtm, thanks for linking it

packages/access-mocsy/package.py
```diff
@@ -6,24 +6,54 @@
...
+            "precision",
+            default="2",
+            description="Precision to use (1 or 2)",
+            values=("1", "2"),
+        )
 
-    # Need: self.spec["mpi"].mpifc
     depends_on("mpi")
 
+    flag_handler = build_system_flags
```
harshula, line 43: After what @edoyango found, should we avoid overriding the default flag handler?

dougiesquire: Let's leave it here for now and we can bulk update all relevant SPRs if that's the way we decide to go

dougiesquire:
Thanks for the review @harshula

PR #266, opened 2025-07-07
265 add am3 executables to path

bschroeter:
As discussed, added the executables to the path.


CodeGat (approved):
Noice, we might have to consider adding this version of the `um` package to `rAM3` in the future.

Why are your commits showing up as unverified?

PR #273, opened 2025-07-11
Work required to add CABLE as library to ESM1p6 build

Whyborn:
Modify UM7 spack package for inclusion of CABLE as library, rather than as source code.


Whyborn:
Yea, will do. There's still work to do to get this bit-wise equivalent to the non-library version, so I'll tidy everything up once I'm at that point.

PR #276, opened 2025-07-21
Link `build-ci 2.0` to `spack-packages`

### LINKED ISSUE #270: Link up `build-ci 2.0` to `spack-packages`
See if we can build modified packages using `build-ci 2.0` on `spack-packages`.

CodeGat:
Closes #270 

## Background

It would be useful to see in a PR if changes to a recipe would actually lead to it being uninstallable. Good thing we have `build-ci 2.0` now! 

This PR adds PR checks to modified package recipes that give results on if the packages are installable with the new changes. 

Furthermore, it also adds a `workflow-dispatch` trigger that allows testing whenever one wants, with different constraints. 

## The PR

* Add a very basic manifest `default.spack.yaml.j2`, that allows packages without specific manifests under `.github/build-ci/manifests/PACKAGE/*spack.yaml.j2` to still be tested in their default state. 
* Add a specific`issm` manifest. 
* Add a new workflow that runs modified packages on `build-ci 2.0`, and on `workflow_dispatch`.

### Pull Request Event

This run: https://github.com/ACCESS-NRI/spack-packages/actions/runs/16436191379 gives an example run in which the `mom5` and `cice5` package recipes each had a (trivial) modification done to them. The Setup CI job successfully found a specific `mom5` manifest under `.github/build-ci/manifests/mom5`, and used the default template for `cice5`. It then sent off the jobs to the self-hosted runners as expected, and the rest is `build-ci 2.0` territory, unaffected by the changes here. See here:

image

### Workflow Dispatch Event

Unfortunately we can't do a `workflow_dispatch` test because it requires the workflow to already be on the default branch to use. I've done a manual test of the commands run instead.


harshula:
Hi @CodeGat , Can there be a default template used by all packages unless there's a specific `spack.yaml` for a package?

CodeGat:
I experimented with a default template (see https://github.com/ACCESS-NRI/spack-packages/issues/270#issuecomment-3095147538) but it seems that the only way to do that would be to add a new input to the workflow that would almost exclusively be used by `spack-packages`, and I'm not sure I want to confuse workflow users. I'll chat with @aidanheerdegen and see if there's a way around it.

CodeGat:
@harshula, @aidanheerdegen, I've managed to get the default templating stuff to work with `spack-packages`. See this commit (now rebased off this branch since it was a test): https://github.com/ACCESS-NRI/spack-packages/pull/276/commits/5f63c551a93e831cdf96a388c1acf77b6c14d6ed, and the associated workflow run: https://github.com/ACCESS-NRI/spack-packages/actions/runs/16436191379 - note that `mom5` uses the specific manifest defined in the package subdirectory, and `cice5` uses the default `template.spack.yaml.j2`.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Dispatch - Get packages
+        id: set-matrix-dispatch
+        if: github.event_name == 'workflow_dispatch'
+        # Either get all packages defined (all subfolders under packages/) or use the input
+        run: |
+          if [[ "${{ inputs.packages-to-test }}" == "ALL" ]]; then
+            echo "pkgs=$(find packages/ -mindepth 1 -maxdepth 1 -type d -printf '%P ')" >> $GITHUB_OUTPUT
+          else
+            echo "pkgs="${{ inputs.packages-to-test }}" >> $GITHUB_OUTPUT
+          fi
```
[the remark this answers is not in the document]
CodeGat: Good point...

.github/build-ci/manifests/template.spack.yaml.j2
```diff
@@ -0,0 +1,6 @@
+spack:
+  specs:
+  - '{{ package }} %intel@2021.10.0'
```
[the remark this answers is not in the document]
CodeGat: Done in https://github.com/ACCESS-NRI/spack-packages/pull/276/commits/4081f359a72d0cb15971823c6dc6a51fd1e89ba1

.github/build-ci/manifests/mom5/spack.yaml.j2
```diff
@@ -0,0 +1,6 @@
+spack:
+  specs:
+  - mom5 %intel@2021.10.0 arch=x86_64
```
aidanheerdegen, line 3:
```suggestion
  - mom5 %{{ intel_compiler }} arch=x86_64
```
Can the non-default manifests access the standard_definitions? If so it should use that I think.

How could a package define it's own data file if there were several manifests they wanted to share common settings across?

If we want to apply `arch` consistently across all tests it should be in the standard_definitions and added to the default I think.

.github/build-ci/manifests/template.spack.yaml.j2
```diff
@@ -0,0 +1,6 @@
+spack:
+  specs:
+  - '{{ package }} %{{ intel_compiler }}'
```
aidanheerdegen, line 3: I thought the reason there was a specific `mom5` package is that the default was going to use `gcc` and `mom5` didn't like that?

Related: I'm thinking the default should have matrix and test both compiler families. To keep it simple it could be a matrix within the manifest.

.github/workflows/ci.yml
aidanheerdegen, line 1: Just have to say this is a remarkably compact bit of workflow logic for quite a complicated outcome. YAY for Build CI2.0.

aidanheerdegen (commented):
I have questions.

.github/build-ci/manifests/mom5/spack.yaml.j2
CodeGat: Yes they can, and in the case of this, they do - it just doesn't use the values it's given because we don't have any of the template vars in the template (yet).

For your second point, it would require some more logic in the `setup` job. We'd have to find a way to associate each matrix entry with a different data file, in a similar way that we associate a manifest with a package. It would be a bit more involved. Currently we associate every matrix entry with the same data file. 

For the last point, fair enough. I think I'll remove the arch for now. It was just to differentiate the template from the "specific" mom5 one. The mom5 manifest is there pretty much as an example for future specific package manifests.

CodeGat: Done in https://github.com/ACCESS-NRI/spack-packages/pull/276/commits/2fa28e51fee796d2b5deb644dd9b4f4b21ad9eee

aidanheerdegen: > The mom5 manifest is there pretty much as an example for future specific package manifests.

Let's use an example we definitely want to do then. ACCESS-ISSM uses `+wrappers` in the build but it is off by default

https://github.com/ACCESS-NRI/access-issm/blob/main/spack.yaml#L9

https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/issm/package.py#L40-L44

and we should also test the `+ad` variant.

Can you make a manifest and test those variants in a matrix within the manifest?

.github/build-ci/manifests/template.spack.yaml.j2
aidanheerdegen: Well I think we should make a decision to test both, see which fail and either create a custom test for those packages to not test both, or fix the package/code to pass ...

CodeGat: Done in https://github.com/ACCESS-NRI/spack-packages/pull/276/commits/0ea183a6bca1d5dc4f6a6102a9891bc94349d027 -  also changed the name of the default template from `template.spack.yaml.j2` to `default.spack.yaml.j2`

PR #281, opened 2025-07-25
access-generic-tracers: Update versions

dougiesquire:
This PR updates versions in the `access-generic-tracers` SPR. Specifically:
- Add version 2025.07.000
- Rename "master" version to "main" (default branch was renamed some months ago). I'm pretty confident this wasn't be used anywhere.
- Add note to delete the "development" version once we're confident it's not being used anywhere.

### DIFF
--- a/packages/access-generic-tracers/package.py
+++ b/packages/access-generic-tracers/package.py
@@ -17,9 +17,9 @@ class AccessGenericTracers(CMakePackage):
 
     maintainers("harshula")
 
-    version("master", branch="master")
-    # TODO: Needs to be changed once changes to build system enter master.
-    version("development", branch="development", preferred=True)
+    version("main", branch="main", preferred=True)
+    # TODO: Delete development version once we're sure we're using main everywhere.
+    version("development", branch="development")
 
     variant(
         "shared",


dougiesquire:
@harshula I'm experiencing a weird issue with the version checksum. The checksum in the SPR was obtained using:

```
$ spack checksum access-generic-tracers
...
version("2025.07.000", sha256="2ff752746ac7645dbe60818fd903b4685fe1f85d9f468adc4b0d50364b7c4274")
```

This checksum matches what I get if I download the tarball and compute the checksum manually
```
$ shasum -a 256 GFDL-generic-tracers-2025.07.000.tar.gz
2ff752746ac7645dbe60818fd903b4685fe1f85d9f468adc4b0d50364b7c4274  GFDL-generic-tracers-2025.07.000.tar.gz
```

However, when I try to install I get:
```
$ spack install access-generic-tracers@2025.07.000 %intel@2021.10.0
...
==> Installing access-generic-tracers-2025.07.000-emcao3k4nv47kivmvcljautlh3dsbsdg [19/19]
==> No binary for access-generic-tracers-2025.07.000-emcao3k4nv47kivmvcljautlh3dsbsdg found: installing from source
==> Fetching https://github.com/ACCESS-NRI/GFDL-generic-tracers/tarball/2025.07.000
==> Error: ChecksumError: sha256 checksum failed for /scratch/tm70/ds0092/tmp/spack-stage/spack-stage-access-generic-tracers-2025.07.000-emcao3k4nv47kivmvcljautlh3dsbsdg/2025.07.000
    Expected 2ff752746ac7645dbe60818fd903b4685fe1f85d9f468adc4b0d50364b7c4274 but got 348ebed7528f4f5eff7647e1d3da7cafb6d5d09854c7bb1cdfa0fc0057ef7a9f. File size = 446848 bytes. Contents = b'\x1f\x8b\x08\x00\x00\x00\x00\x00\x00\x03\xec<ks\xda\xc8...\xf3\xf0<<w\xf0\xfc\xff\xd1\xc8\x89\xb5\x00\x00-\x00'
```

Can you reproduce this behaviour? Any ideas what's going on?

dougiesquire:
> Can you reproduce this behaviour? Any ideas what's going on?

This didn't occur in the build-ci, so I'm going to chalk my issue up to me having an out-of-date git_repos dir or something.

packages/access-generic-tracers/package.py
```diff
@@ -14,12 +14,14 @@ class AccessGenericTracers(CMakePackage):
...
+    url = "https://github.com/ACCESS-NRI/GFDL-generic-tracers/archive/refs/tags/2025.07.000.tar.gz"
 
     maintainers("harshula")
 
-    version("master", branch="master")
-    # TODO: Needs to be changed once changes to build system enter master.
+    version("main", branch="main")
+    # TODO: The development version should be deleted once we're using >=2025.07.000 everywhere.
     version("development", branch="development", preferred=True)
+    version("2025.07.000", sha256="2ff752746ac7645dbe60818fd903b4685fe1f85d9f468adc4b0d50364b7c4274")
```
harshula, line 24: We don't add a new `version()` for each git tag. Instead, the spack.yaml should use `@git.2025.07.000` as the version of access-generic-tracers.

dougiesquire: Thanks @harshula. Is there any harm in keeping this in this case? I've opened #285 to discuss this more generally

harshula: Perhaps focus on the `master` to `main` version renaming in this PR.

dougiesquire: I'd really like to keep this version in unless there's a reason not to?

anton-seaice: We need to be using versions so we can manage the dependencies between versions

anton-seaice, line 24: Using a commit here might be better than a checksum, as it doesn't rely on the existence of the tag/release (which can be deleted) and is still a trusted download method

anton-seaice: https://spack.readthedocs.io/en/latest/packaging_guide_creation.html#git

dougiesquire: Happy to change it. It does means that if a version tag is moved/deleted, that version will still build against the specified commit. If we use checksum of the tag tarball this situation would raise an error. There's an argument that an error is better, but given that we shouldn't be moving/deleting tags it's probably moot...

dougiesquire: Changed in [fa35c8f](https://github.com/ACCESS-NRI/spack-packages/pull/281/commits/fa35c8f693df078f86368721141d1346e70ae305)

anton-seaice: For reasons i can't recall we thought commit was better last time ?

aidanheerdegen: This is a blocker for ESM1.6 development. How can we move this forward?

dougiesquire:
@CodeGat adding `spack.packages.require.all` sections fixed the issue. Are you happy to make this the default for the repo? If so, shall I rebase and remove your commit in this branch once you have?

packages/access-generic-tracers/package.py
dougiesquire: I can remove the `version("2025.07.000",...)` for now if that is the only way to move this forward. We'd really like to get this merged today.

CodeGat:
Yep I'll make it the default, will link the PR when ready

CodeGat:
It's been merged @dougiesquire, you can remove my commit

packages/access-generic-tracers/package.py
dougiesquire: I removed the `version("2025.07.000",...)`

CodeGat (approved):
Verified that it's the `main` branch now rather than `master`, looks good

PR #286, opened 2025-07-29
Split Default Manifests into Intel and GCC

CodeGat:
References #281, thanks @dougiesquire 

## Background

Without a `spack.packages.all.require: %COMPILER` directive, we seem to be mixing compilers, which leads to inconsistent builds. 

In this PR, we split the `default.spack.yaml.j2` into `[intel|gcc].spack.yaml.j2` so we can make use of the `spack.packages.all.require` statements enforcing a particular compiler down the build chain. 

We also update the standard GCC compiler to something a little newer - `13.2.0`

## The PR

- **Split default manifest into intel and gcc with compiler requirements**
- **Update standard gcc compiler to 13.2.0**

## Testing

Testing of the new manifests done in the run referenced by this commit: https://github.com/ACCESS-NRI/spack-packages/pull/281/commits/cb7379882107086480a6ef54fdb89f8250b528e8


.github/build-ci/manifests/intel.spack.yaml.j2
aidanheerdegen, line 1: Will there be a separate oneapi template too at some stage? Or will we just move to that as the default in the future?

aidanheerdegen (dismissed):
LGTM with one small suggestion.

.github/build-ci/manifests/intel.spack.yaml.j2
CodeGat: Possibly! I'm not too fussed on what the default templates are. I've also future-proofed the default manifests bit for this reason, see https://github.com/ACCESS-NRI/spack-packages/pull/286/commits/c2f54ca6af9da6643440d59cae72b86acc626eae

harshula: Separate `oneapi` template, please.

PR #296, opened 2025-08-07
Added openmpi as explicit dependency

bschroeter:
Quick fix heavily inspired by @penguian's gcom example.

Currently running on AM3, which is now the recon phase where it last fell over due to the missing `mpirun` command resulting from an absence of openmpi.

### DIFF
--- a/packages/um/package.py
+++ b/packages/um/package.py
@@ -114,6 +114,11 @@ class Um(Package):
     # used by the FCM configuration of UM.
     depends_on("fcm site=nci-gadi", type="build")
 
+    # Include openmpi directly https://github.com/ACCESS-NRI/spack-packages/issues/293
+    variant("mpi", default=True, description="Build with MPI")
+    depends_on("mpi", when="+mpi", type=("build", "link", "run"))
+    depends_on("openmpi@4.1.7:", when="+mpi^[virtuals=mpi] openmpi", type=("build", "link", "run"))
+
     # For GCOM versions, see
     # https://code.metoffice.gov.uk/trac/gcom/wiki/Gcom_meto_installed_versions
     depends_on("gcom@7.8", when="@:13.0", type=("build", "link"))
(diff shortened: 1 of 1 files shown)


aidanheerdegen:
The CI is failing because we're not do the build checks on `gadi`.

I have made an issue for this, but I don't think it should hold up merging this once @penguian has had a chance to review it.

https://github.com/ACCESS-NRI/spack-packages/issues/298

packages/um/package.py
```diff
@@ -114,6 +114,11 @@ class Um(Package):
     # used by the FCM configuration of UM.
     depends_on("fcm site=nci-gadi", type="build")
 
+    # Include openmpi directly https://github.com/ACCESS-NRI/spack-packages/issues/293
+    variant("mpi", default=True, description="Build with MPI")
+    depends_on("mpi", when="+mpi", type=("build", "link", "run"))
+    depends_on("openmpi@4.1.7:", when="+mpi^[virtuals=mpi] openmpi", type=("build", "link", "run"))
```
aidanheerdegen, line 120: I'd be careful setting this here. `gcom` is already specifying `openmpi@4.1.3:`:

https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/gcom/package.py#L30

Constraints specified at the package level are about making sure the package can compile, and so should very much be minimums, avoiding known conflicts and avoiding known buggy versions.

So I'd match it with what `gcom` does. Effectively it isn't required: once you `depend_on` `gcom` the concretization step will pull in the requirement from `gcom`.

Ultimately the requirement should probably be shifted from `gcom` to this package, but one step at a time, unless you're keen to do it now.

bschroeter: As discussed with @aidanheerdegen just now:

Removing line 120, allowing this to be set at the environment level, which will update the version to 4.1.7 and supersede the 4.1.3 set in `gcom`. Translation to `openmpi` is handled by infra.

bschroeter: @aidanheerdegen, see build here: https://github.com/ACCESS-NRI/ACCESS-AM3/actions/runs/16819521835/job/47643574994

4.1.7 is being picked up and used.

aidanheerdegen (approved):
LGTM.

PR #302, opened 2025-08-14
Fix the um jules_sources and um_sources variants

### LINKED ISSUE #299: UM Spack package clobbers the jules_sources and um_sources variants
The current UM Spack package clobbers the `jules_sources` and `um_sources` variants with the default value `"NA"` unless these values are specified on the command line. The values set in `model/*/rose-app.conf` are not used. This is incorrect for all model variants except possibly for `vn13p1-am`.

See https://github.com/ACCESS-NRI/spack-packages/pull/274 and especially https://github.com/ACCESS-NRI/spack-packages/pull/274#discussion_r2238884097

In https://github.com/ACCESS-NRI/spack-packages/blob/2025.08.000/packages/um/package.py as well as [the current main branch](https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/um/package.py), two variants are defined twice in `class Um`: `jules_sources` and `um_sources`:

- `jules_sources` is first defined at line 111 with default `"none"`: 
https://github.com/ACCESS-NRI/spack-packages/blob/02ce16f8793db740c3979edcbdfd665d68d56b19/packages/um/package.py#L111
then defined again at line 162 with default `"NA"`:
https://github.com/ACCESS-NRI/spack-packages/blob/02ce16f8793db740c3979edcbdfd665d68d56b19/packages/um/package.py#L162
- `um_sources` is first defined at line 111 with default `"none"`: 
https://github.com/ACCESS-NRI/spack-packages/blob/02ce16f8793db740c3979edcbdfd665d68d56b19/packages/um/package.py#L111
then defined again at line 166 with default `"NA"`:
https://github.com/ACCESS-NRI/spack-packages/blob/02ce16f8793db740c3979edcbdfd665d68d56b19/packages/um/package.py#L166

The later definitions supersede the earlier ones. See https://spack.readthedocs.io/en/latest/_modules/spack/variant.html This has the effect of defining these variants as follows:

```
    jules_git_url = "https://github.com/ACCESS-NRI/JULES.git"
    variant("jules_sources", default="NA", values=str, multi=False, description=f"Optional JULES sources branch/tag/commit from {jules_git_url}.")
    um_git_url = "https://github.com/ACCESS-NRI/UM.git"
    variant("um_sources", default="NA", values=str, multi=False, description=f"Optional UM sources branch/tag/commit from {um_git_url}.")
```
The `setup_build_environment` module at
https://github.com/ACCESS-NRI/spack-packages/blob/02ce16f8793db740c3979edcbdfd665d68d56b19/packages/um/package.py#L204
is intended to create environment variables corresponding to variants, by parsing the `rose-app.conf` file corresponding to `model`, and then using the `spec` value to override the `config_env` value, if the `spec` value is not the default `"none"`.

The `"jules_sources"` and `"um_sources"` variants are contained in `self._other_variants` and therefore at line 281, their value is compared to `"none"`.
https://github.com/ACCESS-NRI/spack-packages/blob/02ce16f8793db740c3979edcbdfd665d68d56b19/packages/um/package.py#L281 Since the default value is now `"NA"`, this value is used instead of whatever is defined in the model `rose-app.conf` file. For every model other than possibly for `vn13p1-am`, this is incorrect.

A quick fix for all other models would be to remove the second variant definitions for `"jules_sources"` and `"um_sources"` , but the behaviour for `vn13p1-am` may also need to be reconsidered.

penguian:
Closes #299 
I looked at how the `jules_sources` and `um_sources` variants were being used and I now agree with @bschroeter that new variants are needed. 

The changes add the `jules_tag` and `um_tag` variants and the `_resource_cfg` dictionary. The tag variants are processed in loops, using the `_resource_cfg` dictionary. The user is warned when variant values conflict and are replaced, and also when tag variants are ignored. 

The package now uses SSH Github URLs to prevent the `git clone` from asking for a login. This should work for all users that have correct SSH key setup for Github and have access to the private repositories used here.

Successfully tested with the `vn13p1-am` and `vn13p5-rns` models. Reviewers could repeat these tests:

1.
```
spack install um@13.1 ~eccodes model=vn13p1-am jules_tag=wrong um_tag=AM3-dev %intel@2021.10.0
```
This will print info about the `jules_tag`, `jules_sources`, `um_tag` and `um_sources` variants, then run the `git clone` for `jules`, but fail in `git checkout` with:
```
error: pathspec 'wrong' did not match any file(s) known to git
==> patch() function failed for um
==> Error: ProcessError: Command exited with status 1:
    '/bin/git' 'checkout' 'wrong'
```
2.
```
$ spack install um@13.1 ~eccodes model=vn13p1-am jules_tag=AM3-dev um_tag=AM3-dev %intel@2021.10.0
```
This will print info about the `jules_tag`, `jules_sources`, `um_tag` and `um_sources` variants, run `git clone` and `git checkout` for JULES and UM, then successfully build and install.

3.
```
$ spack install um@13.1 ~eccodes model=vn13p1-am %intel@2021.10.0
```
This will not print warnings about the missing tag variants. The build will start, but will fail with
```
==> Error: ProcessError: Command exited with status 2:
    '.../ACCESS-NRI/release/.../fcm-2021.05.0-.../bin/fcm' 'make' '--new' '--config-file=.../ACCESS-NRI/spack-packages/packages/um/fcm-make.cfg' '--directory=.../tmp/spack-stage/spack-stage-um-13.1-.../spack-src/../spack-build' '--jobs=4'
See build log for details:
  .../tmp/spack-stage/spack-stage-um-13.1-.../spack-build-out.txt

```
where the build log contains
```
...
[FAIL] cable_jules_links_mod.mod: bad or missing dependency (type=1.include)
...
```
because the underlying `vn13p1-rns` configuration points to source code on the MOSRS mirror that is inconsistent.

4. 
```
$ spack install um@13.5 +eccodes model=vn13p5-rns um_tag=AM3-dev um_sources=nonesuch %intel@2021.10.0
```
This will warn about the `um_tag` and `um_sources` variants, start the build, and fail with:
```
     33    [FAIL] file:///g/data/ki32/mosrs/um/main/nonesuch: not found
  >> 34    [FAIL] svn: warning: W170000: URL 'file:///g/data/ki32/mosrs/um/main/nonesuch' non-existent in revision 130799
```
5.
```
$ spack install um@13.5 ~DR_HOOK model=vn13p5-rns %intel@2021.10.0
```
This will successfully build and install.


bschroeter (approved):
LGTM

penguian:
@bschroeter I am also updating [packages/um/model/vn13p1-am/rose-app.conf](https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/um/model/vn13p1-am/rose-app.conf) to match the changes in https://github.com/ACCESS-NRI/access-am3-configs/blob/main/app/fcm_make_um/rose-app.conf

This will need to be tested in the [AM3](https://github.com/ACCESS-NRI/ACCESS-AM3).

penguian:
Notes:

1. For `vn13p1-am` if `ukca_ref` is not `"none"` then [line 371 of `package.py`](https://github.com/ACCESS-NRI/spack-packages/blob/88ed72fa72e5be36a122acb489eefaa046222fa3/packages/um/package.py#L371) will clobber whatever is in `ukca_sources` with the location of the Github clone obtained from `resource_path`. 
@bschroeter Should the code just prepend `resource_path` to `ukca_sources`? What is the requirement here?

1. At [line 22 of `model/vn13p1-am/rose-app.conf`](https://github.com/ACCESS-NRI/spack-packages/blob/88ed72fa72e5be36a122acb489eefaa046222fa3/packages/um/model/vn13p1-am/rose-app.conf#L22) the definition
```
config_root_path=fcm:um.xm_br/dev/jhansrbinovsky/u-cv185
```
is kept. To change this to point to the `resource_path` for UM requires more changes to `package.py`. @bschroeter Do you want these extra changes in this branch and pull request, or do you want to create a separate issue, setting out the requirements for `config_root_path`?

packages/um/package.py
```diff
@@ -329,26 +395,30 @@ def _build_dir(self):
...
+        # This patch is relevant only for models that use Github URLs.
+        # Only one model so far, but this may change in future.
+        spec = self.spec
+        model = spec.variants["model"].value
+        if model == "vn13p1-am":
+            # Get the root to the resources
+            resources_root = join_path(self.stage.source_path, "resources")
+
+            # Optional sources (i.e. AM3)
+            for var in self._resource_cfg:
```
harshula, line 411: Perhaps use the same variable name for iterating over `self._resource_cfg` as you have done elsewhere.

packages/um/package.py
```diff
@@ -329,26 +395,30 @@ def _build_dir(self):
...
+                spec_value = spec.variants[var].value
```
harshula, line 412: Perhaps use `ref_value` instead of `spec_value` as you have done elsewhere.

packages/um/package.py
```diff
@@ -225,6 +233,44 @@ def check_model_vs_spec(model, config_env, var, spec_value):
...
+            """
+            Check the values set by the variants sources_var and ref_var
+            against any existing sources value in config_env, and remind
+            that the sources value will be overridden by resource_path.
+            """
+            sources_value = spec.variants[sources_var].value
+            ref_value = spec.variants[ref_var].value
+            tty.info(f"The spec sets {ref_var}={ref_value}")
+            if sources_value == "none":
+                # In this case, the spec has not overridden any value
```
harshula, line 251: "any value"? Or has not overridden `sources_var`'s value?

packages/um/package.py
```diff
@@ -225,6 +233,44 @@ def check_model_vs_spec(model, config_env, var, spec_value):
...
+                    env_value = config_env[sources_var]
+                    if env_value == "":
+                        tty.info(
+                            f"The {model} model sets {sources_var}=''.")
+                    else:
+                        tty.warn(
+                            f"The {model} model sets "
+                            f"{sources_var}={env_value}.")
+            else:  # sources_value != "none"
+                # In this case, the spec has already overridden any value
```
harshula, line 266: "any value"? Or has  overridden sources_var's value?

packages/um/package.py
```diff
@@ -225,6 +233,44 @@ def check_model_vs_spec(model, config_env, var, spec_value):
...
+                # in the model configuration.
+                assert sources_value == config_env[sources_var]
```
harshula, line 268: Is this true because of the code in line 324 to 329?

harshula, line 268: Is this true because of the code in lines 324 to 329?

penguian: Yes.

packages/um/package.py
penguian: I changed the comment. Hopefully it is clearer.

packages/um/package.py
penguian: I changed the comment. Hopefully it is clearer.

penguian:
Merging.

PR #308, opened 2025-08-25
access-generic-tracers: add 2025.04.001

dougiesquire:
This PR adds `access-generic-tracers@2025.04.001`. `2025.04.001` is the same as the `2025.04.0` release tag, but with the CMake build system backported so that it can be built with the latest version of the SPR.


anton-seaice (approved):
Looks good to me, CI builds succesfully, hash matches tag from https://github.com/ACCESS-NRI/GFDL-generic-tracers/releases/tag/2025.04.001

PR #315, opened 2025-08-28
Do not use -V in fflags: it fools the Spack compiler wrapper

penguian:
Remove the `-V` flags from the `fcm build` config within the UM7 Spack package recipe.

[At lines 277-289 of the Spack compiler wrapper](https://github.com/ACCESS-NRI/spack/blob/99f79c7cf1c97c55f5d1920679a4d5bb998b5dc3/lib/spack/env/cc#L277) there is a test to see if there are any arguments that represent a version check. In this case, [at lines 367-369](https://github.com/ACCESS-NRI/spack/blob/99f79c7cf1c97c55f5d1920679a4d5bb998b5dc3/lib/spack/env/cc#L367) the wrapper just executes the compiler command without adding arguments from `$SPACK_FFLAGS`.
```
# If any of the arguments below are present, then the mode is vcheck.
# In vcheck mode, nothing is added in terms of extra search paths or
# libraries.
if [ -z "$mode" ] || [ "$mode" = ld ]; then
    for arg in "$@"; do
        case $arg in
            -v|-V|--version|-dumpversion)
                mode=vcheck
                break
                ;;
        esac
    done
fi
[...]
if [ "$mode" = vcheck ]; then
    exec "${command}" "$@"
fi
```
Because lines 182, 186 and 188 of the FCM build config in [packages/um7/package.py](https://github.com/ACCESS-NRI/spack-packages/blob/main/packages/um7/package.py) contain `"-V"` flags, when `spack install` is run for `um7`, the Spack compiler wrapper is ignoring the value of `fflags`. The fix for now is simply to remove these flags, as they are not needed, and do not affect the code generated by the Intel Fortran compiler.

See also [the documentation for the Intel Fortran compiler `-logo` aka `-V` option](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2023-0/logo.html).

### DIFF
--- a/packages/um7/package.py
+++ b/packages/um7/package.py
@@ -179,13 +179,13 @@ tool::cpp                                          cpp
 tool::cppflags
 tool::cppkeys                                      {CPPKEYS}
 tool::fc                                           mpif90
-tool::fflags                                       {FO}  -g   -traceback  {FDEBUG} -V -i8 -r8      -fp-model precise {FFLAGS}
-tool::fflags::control::coupling::dump_received     {FO} {FG} {FTRACEBACK} {FDEBUG}            -mp1 -fp-model strict  {FFLAGS}
-tool::fflags::control::coupling::dump_sent         {FO} {FG} {FTRACEBACK} {FDEBUG}            -mp1 -fp-model strict  {FFLAGS}
-tool::fflags::control::coupling::oasis3_atmos_init {FO} {FG} {FTRACEBACK} {FDEBUG}    -i4 -r8 -mp1 -fp-model strict  {FFLAGS}
-tool::fflags::control::top_level::atm_step         -O0   -g  {FTRACEBACK} {FDEBUG} -V -i8 -r8 -mp1 -fp-model strict  {FFLAGS}
-tool::fflags::control::top_level::set_atm_pointers -O0   -g   -traceback  {FDEBUG}    -i8 -r8      -fp-model strict -ftz -std95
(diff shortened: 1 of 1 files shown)


Whyborn (approved):

manodeep:
This "feature" seems bad design - the idea of making the mode as `vcheck` makes sense if `-v/-V` etc is the **only** argument, not when it matches *any* of the arguments. This should be reported as an issue to upstream - what do y'all think?

penguian:
> This "feature" seems bad design - the idea of making the mode as `vcheck` makes sense if `-v/-V` etc is the **only** argument, not when it matches _any_ of the arguments. This should be reported as an issue to upstream - what do y'all think?

The `vcheck` should definitely be reported upstream, but note that as far as I can tell, this script is no longer used in Spack 1.X.

penguian:
> I think it may be useful to have the version of the compiler printed to the logs- could we have `-logo` added to `{FDEBUG}`? Do you know why only some components had `-V` applied in the first place?

The original code is at https://github.com/ACCESS-NRI/cmip5/blob/dev/jxs599/trunk_ESM1.5/submodels/UM/ummodel_hg3/cfg/bld-hadgem3-mct.cfg
I don't know why only some sources have `-V` and others do not. Jhan Srbinovsky may know.

manodeep:
The CI failure seems to be related to the [private GCOM4 repo issue](https://github.com/ACCESS-NRI/spack-packages/actions/runs/17287571815/job/49067594258?pr=315#step:13:4154) (pinging @CodeGat) and unrelated to the changes in this PR. 

Anything else holding up merging this PR?

CodeGat:
Ah, thanks @manodeep. I'll fix this up quickly.

penguian:
Merging.

PR #317, opened 2025-09-02
Copy `from spack.package import *` from upstream template

harshula:
This is the first commit in the `api-v2` branch. Since it is compatible with Spack `v0.22` and `v1.0`, we might as well merge it into `main`.

### DIFF
--- a/packages/access-fms/package.py
+++ b/packages/access-fms/package.py
@@ -5,7 +5,8 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-from spack.package import find_headers, find_libraries
+from spack.package import *
+
 
 # Based on upstream $spack/var/spack/repos/builtin/packages/fms/package.py
 class AccessFms(CMakePackage):
--- a/packages/access-generic-tracers/package.py
+++ b/packages/access-generic-tracers/package.py
@@ -3,7 +3,8 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-from spack.package import find_libraries
+from spack.package import *
+
 
 class AccessGenericTracers(CMakePackage):
     """This is a fork of the NOAA-GFDL/ocean_BGC repo managed by NOAA-GFDL.
--- a/packages/access-issm/package.py
+++ b/packages/access-issm/package.py
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-from spack.package import BundlePackage, maintainers, version, depends_on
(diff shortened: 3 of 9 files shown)


aidanheerdegen:
That is a lot of failing checks. Do we know why?

CodeGat:
Yes, this is because the Docker image used is out of date (it still has the `user` scope disabled due to building it with `spack-enable.bash`). I'd pushed an image with the `user` scope enabled, but that was for the `v1.0` image. I'd forgotten to do the `v0.22` image, but it is building now. It will be fixed by the end of today. Will comment when it's fixed.

CodeGat:
* `ACCESS-rAM3` failures are due to the `um` package attempting to install from the Gadi repository, so I'm thinking about just disabling tests for any packages that rely on the `um` package. Alternatively it serves as a good reminder to get it fixed...
* `ACCESS-OM3` had a read timeout, which seems like a gremlin that should go away next run: https://github.com/ACCESS-NRI/spack-packages/actions/runs/17404574596/job/49418264701?pr=317#step:14:2007
* `CICE4` looks like a legit error: https://github.com/ACCESS-NRI/spack-packages/actions/runs/17404574596/job/49424087915?pr=317#step:14:5404

And the syntax checks are normal until they are fixed

CodeGat:
Right. In which case we should remove the `gcc` case by copying `.github/build-ci/manifests/intel.spack.yaml.j2` to `a new `.github/build-ci/manifests/cice4/` folder - so it runs only the intel build. 

As for the `um` related discussion, do you think we should exclude CI testing for all packages that use it, due to the reliance on Gadi to build the package? Or is it a good reminder to get it fixed?

aidanheerdegen:
> As for the `um` related discussion, do you think we should exclude CI testing for all packages that use it, due to the reliance on Gadi to build the package? Or is it a good reminder to get it fixed?

I'm  torn. Fixing it is not simple, but maybe if we don't have things like this prodding us maybe we won't make the necessary effort.

harshula:
Hi @aidanheerdegen & @CodeGat , The changes you are discussing are not going to be implemented in this PR. Please create a separate PR/Issue for that discussion.

aidanheerdegen:
@CodeGat has a [fix for the strange manifest paths](https://github.com/ACCESS-NRI/spack-packages/pull/319#event-19473851326). Should @harshula rebase this PR on to that @CodeGat?

CodeGat:
In this case, the tests will still fail due to the `um` stuff. I think it's fine to merge as is.

aidanheerdegen (approved):

PR #319, opened 2025-09-03
`build-ci`: Fix incorrect manifest search

CodeGat:
See run https://github.com/ACCESS-NRI/spack-packages/actions/runs/17404574596/job/49459311901?pr=317 (specifically, the incorrect `.github/build-ci/manifests/access-om3/gcc.spack.yaml.j2` for `access-ram3` - it should either be a `access-ram3` manifest or the default `.github/build-ci/manifests/gcc.spack.yaml.j2`). 

## The PR

* `unset` the `manifest_paths` variable between loops, as the `[ -z "$manifest_paths" ]` was picking up the previous loop iterations value. Whoops!


aidanheerdegen (approved):
Good catch

PR #332, opened 2025-10-08
SPR modifications to support api-v2 (Spack v1.x)

harshula:
DRAFT to track changes


CodeGat:
Hey @harshula, the error is because of the changed directory structure for `spack-packages` between `api-v1` and `api-v2`. Unfortunately I can't suggest changes directly because the bits I want to change are unmodified. But:

In https://github.com/ACCESS-NRI/spack-packages/blob/cc4ec27340bc179b1efade7c74938e49ea61d5b6/.github/workflows/ci.yml#L5, you need to change that to `- spack_repo/access/nri/packages/**`

In https://github.com/ACCESS-NRI/spack-packages/blob/cc4ec27340bc179b1efade7c74938e49ea61d5b6/.github/workflows/ci.yml#L51, change it to `path: ./spack_repo/access/nri/packages`

CodeGat:
Just noting that I've updated the upstream image in the cluster to `:rocky-v1.0-2025.11.000`, which contains the `intel-oneapi-compilers-classic@2021.10.0 target=x86_64` compiler.

CodeGat:
@harshula, once this run is done, feel free to up the `max-parallel` to `7`

CodeGat:
I've updated the image in the infrastructure to `:rocky-v1.1-2025.12.000-test`, please use that as the `container-image-version`. Furthermore, that image is only compatible with the new version of `build-ci` currently being tested, so it might be good to knock two birds with one stone.

CodeGat:
Hi @harshula, it was not - I'll recreate the image without the compilers and update the infra!

CodeGat:
There was a bug in `build-ci` which led to the wrong upstream being selected from the kubernetes cluster. This has now been fixed, and would explain the mismatch between the hashes found by the runner in the upstream (which actually came from a `0.22` upstream) and the hashes in the actual upstream needed (the `1.X` one).

CodeGat:
So you're saying...that's that? It works? :astonished:

CodeGat (approved):
I'm going to pre-emptively approve this even though the checks haven't finished, that's how hopeful I am. Even though this branch will be the next default branch.

harshula:
### Branch main is deprecated and will be deleted in the future.

This PR does NOT need to be merged into the `main` branch. We will be using `api-v2` branch with Spack v1.

PR #338, opened 2025-10-21
libaccessom2: specify the virtual rather than the concrete package

harshula:
$ spack providers pkgconfig
pkgconfig:
pkg-config  pkgconf


harshula:
This change is required for building `access-om` on Setonix because they build and install `pkg-config` and not `pkgconf`.

PR #340, opened 2025-10-26
Changes required to use CABLE as a library for ESM1.6

Whyborn:
Associated PR in ESM1.6 that shows bitwise compatibility with non-library build: https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/156
UM PR that removes the CABLE code: https://github.com/ACCESS-NRI/UM7/pull/176

The implementation in the UM7 spack package feels quite clunky but I couldn't think of a better way to do it. Happy to hear suggestions on how to improve it.

### DIFF
--- a/packages/cable/package.py
+++ b/packages/cable/package.py
@@ -35,7 +35,7 @@ class Cable(CMakePackage):
         default="none",
         values=(
             "none",
-            "esm1.6"
+            "ESM1.6"
         ),
         description="Build CABLE science library object.",
     )
--- a/packages/cable/package.py
+++ b/packages/cable/package.py
@@ -34,7 +34,7 @@ class Cable(CMakePackage):
         "library",
         default="none",
         values=(
-            conditional("ESM1.6", when="@access-esm1.6")
+            "ESM1.6"
         ),
         description="Build CABLE science library object.",
     )
--- a/packages/cable/package.py
+++ b/packages/cable/package.py
@@ -17,7 +17,7 @@ class Cable(CMakePackage):
     homepage = "https://github.com/CABLE-LSM/CABLE"
     git = "https://github.com/CABLE-LSM/CABLE.git"
 
-    maintainers("SeanBryan51")
+    maintainers("SeanBryan51", "Whyborn")
 
     version("main", branch="main")
 
--- a/packages/cable/package.py
+++ b/packages/cable/package.py
@@ -33,6 +33,9 @@ class Cable(CMakePackage):
     variant(
         "library",
         default=False,
+        values=(
+            conditional("ESM1.6", when="@access-esm1.6")
+        ),
         description="Build CABLE science library object.",
     )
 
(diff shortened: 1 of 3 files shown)


Whyborn:
@Harshula Could I get your assistance debugging this failure: I'm trying to work out the CABLE CI failure- says `cannot satisfy cable@access-esm1.6`, but it builds with no problems in [this ESM1.6 PR](https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/156) which uses that spec. I'm expecting the UM7 build failure, as the CI will be trying to build with CABLE as library while still having it as source code, UM7 PR is [here](https://github.com/ACCESS-NRI/UM7/pull/176).

Whyborn:
@penguian as per the conversation in the ESM1.6 meeting this morning, the GCC build failures should not block merging. Further, we've removed the GCC build from UM7's build CI.

packages/um7/package.py
```diff
@@ -136,6 +143,53 @@ def edit(self, spec, prefix):
...
+        # source code.
+        CABLE_excl_deps = ""
+        with when("@access-esm1.6"):
+            CABLE_excl_deps = """
+excl_dep                                           USE::cable_def_types_mod
+excl_dep                                           USE::cbl_masks_mod
+excl_dep                                           USE::cable_other_constants_mod
+excl_dep                                           USE::cable_math_constants_mod
+excl_dep                                           USE::cable_phys_constants_mod
+excl_dep                                           USE::cable_soil_params_mod
```
SeanBryan51: Just a suggestion. Typically libraries expose their interface through a single module, have you considered having a `cable_mod` which contains all CABLE specific procedures/data structures? This could improve the organisation (and a single `excl_dep USE::cable_mod` 🙂 )

SeanBryan51 (dismissed):
Approving, no need to address my suggestion to expose CABLE specifics through a single module in this PR since its a pretty big change. But it would be great if that was put in later IMO.

packages/um7/package.py
Whyborn: Yea Claire and I discussed this, and it would be nice, but a future task and not worth stopping the library work for now.

packages/cable/package.py
```diff
@@ -32,7 +32,11 @@ def url_for_version(self, version):
 
     variant(
         "library",
-        default=False,
+        default="none",
+        values=(
+            "none",
+            "ESM1.6"
```
penguian: I would prefer the value "access-esm1.6" to be used, to match the other packages, unless there is a good reason to use "ESM1.6".

penguian (changes requested):
Try changing ESM1.6 to access-esm1.6

penguian (approved):
Looks OK to me.

harshula:
Hi @Whyborn & @penguian , 

re: https://github.com/ACCESS-NRI/access-spack-packages/pull/372

@penguian did extensive testing of `with when()`. We should only use it in the member variable declaration area of the class for conditional `depends_on()` calls.

PR #346, opened 2025-11-03
CI: Underscore to hyphen fix, small readability changes

CodeGat:
## Background

Currently in https://github.com/ACCESS-NRI/spack-packages/pull/332, packages are being tested for compliance for changes in `spack v1`. 
One of those changes is that the directory structure has changed, and package directories should have `_` rather than `-` (so they are valid python modules). They are then converted back into their `-`-style packages for `spack`.

However, this means that the CI has been failing, because it has been using the `_` package directories as package names, which doesn't work.

I've also added some readability fixes to the matrix, so the name of the package comes first, rather than the filepath. 

## The PR

* Replace `_` from package directory name with `-` for package names in `template_value`
* Swap `filepath` and `template_value` in matrix, so the package name is more visible in the CI on the web
* Also update paths to `api-v2` format (I had missed the `workflow_dispatch` case!)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ jobs:
       DEFAULT_SPACK_REF: 'releases/v1.0'
       DEFAULT_SPACK_CONFIG_REF: 'main'
       DEFAULT_SPACK_PACKAGES_REF: ${{ github.event.pull_request.head.sha }}
+      PACKAGES_ROOT_DIR: spack_repo/access/nri/packages
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
       spack-packages-ref: ${{ steps.set-refs.outputs.spack-packages-ref }}
@@ -45,10 +46,10 @@ jobs:
         if: github.event_name == 'pull_request'
         uses: tj-actions/changed-files@ed68ef82c095e0d48ec87eccea555d944a631a4c  # v46.0.5
         with:
-          # We only care about the directory names under packages/ - since they are also the package names
+          # We only care about the directory names under the env.PACKAGES_ROOT_DIR - since they are also the package names
           # and we want to find changes to the package.py files in those directories
           dir_names: true
-          path: ./spack_repo/access/nri/packages
+          path: ./${{ env.PACKAGES_ROOT_DIR }}
           files: '*/package.py'
 
       - name: Dispatch - Get packages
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -57,7 +57,7 @@ jobs:
         # Either get all packages defined (all subfolders under packages/) or use the input
         run: |
           if [[ "${{ inputs.packages-to-test }}" == "ALL" ]]; then
-            pkgs=$(find packages/ -mindepth 1 -maxdepth 1 -type d -printf '%P ')
+            pkgs=$(find spack_repo/access/nri/packages/ -mindepth 1 -maxdepth 1 -type d -printf '%P ')
```
harshula: Is it possible for the path not to be hardcoded?

.github/workflows/ci.yml
```diff
@@ -84,17 +84,18 @@ jobs:
         # Get file paths to manifests for each package, and a template, then finally
         # Convert into a JSON array for the matrix, one of the form:
         # [
-        #   {"filepath": ".github/build-ci/manifests/mom5/intel.spack.yaml.j2", "template_value": "mom5"},
-        #   {"filepath": ".github/build-ci/manifests/mom5/amd.spack.yaml.j2", "template_value": "mom5"},
-        #   {"filepath": ".github/build-ci/manifests/cice5/spack.yaml.j2", "template_value": "cice5"},
+        #   {"template_value": "mom5", "filepath": ".github/build-ci/manifests/mom5/intel.spack.yaml.j2"},
+        #   {"template_value": "mom5", "filepath": ".github/build-ci/manifests/mom5/amd.spack.yaml.j2"},
+        #   {"template_value": "cice5", "filepath": ".github/build-ci/manifests/cice5/spack.yaml.j2"},
```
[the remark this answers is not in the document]
CodeGat: This is so it is a bit easier to understand what is being tested in the Web UI. Currently, it looks like:
image
Instead of a filepath to the manifest as the first element, it will be the package name, which should make it a little bit easier to determine what was tested in big matrices like the ones you're testing.

.github/workflows/ci.yml
CodeGat: Not that I can think of...it's not like we can access `spack` to get something like `spack location --repo access_spack_packages`. Since this CI will be sticking around on the `api-v2` branch, which has this structure, I think it's okay for this to remain hardcoded.

harshula: Is there any way to receive the path as an input? e.g. From a config file or env variable? If that's not possible, can you define the `path` as a common variable instead of in multiple locations: https://github.com/ACCESS-NRI/spack-packages/blob/c98325d188972862f87069baa4c92aefba399262/.github/workflows/ci.yml#L43-L63

CodeGat: Done in https://github.com/ACCESS-NRI/spack-packages/pull/346/commits/9f9900ad3b8b974b1ebc6b4da7ffc54377210597

harshula: I'm doing test builds in PR 332. Let's see how the underscore to hyphen mapping will go.

.github/workflows/ci.yml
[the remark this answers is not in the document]
CodeGat: Hey @harshula, unfortunately not. That section of the workflow is raw string only and can't evaluate `env`s (or any templatable stuff, actually). That bit of duplication is here to stay :cry:

PR #349, opened 2025-11-06
Update ISSM package

lawrenceabird:
This PR contains the following main updates:
- Add `access-release` as preferred branch on [ACCESS-NRI ISSM fork](https://github.com/ACCESS-NRI/ISSM/tree/access-release) to handle official releases.
- Fix `+ad` variant by adjusting conditional syntax for variants.
- Remove Python version dependency here as this is specified in the `spack.yaml`.

### DIFF
--- a/packages/issm/package.py
+++ b/packages/issm/package.py
@@ -69,15 +69,13 @@ class Issm(AutotoolsPackage):
     # --------------------------------------------------------------------
     # Dependencies
     # --------------------------------------------------------------------
-
-    # Build-time & runtime dependencies
-    # --------------------------------------------------------------------
+    # Build-time tools
     depends_on("autoconf", type="build")
     depends_on("automake", type="build")
     depends_on("libtool", type="build")
     depends_on("m4", type="build")
 
-    # Runtime MPI dependency
+    # Core build + runtime deps
     depends_on("mpi")
 
     # Conditional dependencies


packages/issm/package.py
```diff
@@ -83,27 +79,28 @@ class Issm(AutotoolsPackage):
...
+    with when("+ad"):
+        depends_on("metis")
+        depends_on("parmetis")
+        depends_on("mumps~openmp", when="~openmp")
+        depends_on("mumps+openmp", when="+openmp")
+        depends_on("scalapack")
+        depends_on("codipack")
+        depends_on("medipack")
+
+    # Other dependencies
```
harshula: Have we received an answer from ISSM upstream to the question, "Which libraries are required to build ISSM with Automatic Differentiation support? Which of these libraries should NOT be used for a default ISSM build?"

lawrenceabird: Yes -- AD can now be compiled with `petsc`, but it requires `adjointpetsc` as well. This is not readily available in existing spack packages. The alternate installation method is this (using `codipack` and `medipack`, and no `petsc`) for AD installations.

harshula: Can you please add a comment above the dependencies with the response from ISSM upstream?

packages/issm/package.py
```diff
@@ -82,28 +78,30 @@ class Issm(AutotoolsPackage):
...
-    # Note: ISSM's MUMPS support is not compatible with the Spack-provided
-    # MUMPS, so we use the one provided by the ISSM team.
-
-    # Optimiser
+    # When building "default" ISSM, use Petsc (including metis [incl. parmetis], mumps, and scalapack)
+    with when("~ad"):
+        depends_on("petsc~examples+metis+mumps+scalapack")
+
+    # When building with AD support, do not use Petsc; instead use CoDiPack + MeDiPack.
+    # Other dependencies (metis, parmetis, mumps, scalapack) remain.
```
harshula: re: "Other dependencies [...] remain"
Does "remain" mean those dependencies  should be there for the default ISSM build?

lawrenceabird: Do you mean just syntactically, or functionally? Using the `petsc` variants includes these `depends_on(XX)` calls internally, so does that not suggest that these dependencies are inherent?

harshula: Do you believe that the aforementioned two examples are functionally equivalent from the perspective of a package manager, like Spack?

harshula: From the perspective of a package manager, like Spack, they are not functionally equivalent.

1. `depends_on("petsc+metis+mumps+scalapack")` - It tells the package manager that ISSM requires `petsc` and that `petsc` must be built with the 3 variants enabled for ISSM to be built correctly.

2. The following tells the package manager that ISSM requires `petsc`, `metis`, `mumps` and `scalapack` to build correctly:
```
depends_on("petsc")
depends_on("metis")
depends_on("mumps")
depends_on("scalapack")
```

lawrenceabird: Thanks for this explanation, @harshula . I am pinging @MathieuMorlighem here -- he'll be able to clarify these requirements for ISSM.

@MathieuMorlighem -- Please see the discussion above RE: dependencies for building ISSM. Without explicitly specifying `metis`, `mumps`, and `scalapack` as dependencies required for ISSM results in a successful build. My understanding is that the approach of specifying `depends_on("petsc~examples+metis+mumps+scalapack")` is equivalent to how we install `petsc` using the `$ISSM_DIR/externalpackages/petsc/install-XXX.sh` scripts (i.e., with these variants), and should be sufficient here. Can you confirm whether these additional dependencies are explicitly needed to build ISSM?
 
@harshula -- please clarify anything here that I'm misrepresenting.

Thanks both.

harshula: Hi @MathieuMorlighem , Thanks for helping us with this. 

For the non-AD build of ISSM, if the following is true:
* `PETSc` is built with `metis`
* ISSM is built with the aforementioned `PETSc`

Can ISSM be built *without* `metis` and still "run ISSM in parallel"?

lawrenceabird: Thanks @MathieuMorlighem!

Hi @harshula -- the way that I interpret this is that if we use `petsc` with the variants that we have specified, we don't require the other individual packages. Everything that we need is handled by `petsc` given these variants. Does that seem reasonable?

lawrenceabird: Hi @harshula -- can you explain a little more about what you mean here? These arguments are intentionally not conditional. They are needed for both `+ad` and non-AD builds.

packages/issm/package.py
```diff
@@ -82,28 +78,30 @@ class Issm(AutotoolsPackage):
...
+    with when("+ad"):
+        depends_on("metis")
+        depends_on("parmetis")
+        depends_on("mumps~openmp", when="~openmp")
+        depends_on("mumps+openmp", when="+openmp")
+        depends_on("scalapack")
```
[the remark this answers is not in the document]
harshula: Hi @lawrenceabird , From the Spack package manager's perspective there is a functional difference in its dependency graph. Without your recent change, `metis` appears as a dependency of `issm`. After your change, `metis` no longer appears on the dependency graph as a dependency of `issm`, but of `PETSc`.

It's important to be aware that a `variant` is not synonymous with a `dependency`. Unbeknown to the `ISSM` SPR, setting variants of the `PETSc` SPR happens to add `PETSc` dependencies that are also needed by `ISSM`. The correct way to do this is for `ISSM` to unconditionally depend on Spack packages that it unconditionally uses to index a data structure to obtain the prefix and pass it as a configure argument.

lawrenceabird:
Hi @harshula:

I have made this change (see [here](https://github.com/ACCESS-NRI/access-spack-packages/pull/349/commits/3cd195042451007904757efb93e6f5b3e5626b12)). However, when I test building ISSM with this change, the default (`~ad`) does not appear to be respected and `codipack` and `medipack` are used rather than `petsc`. See the concretization step [here](https://github.com/ACCESS-NRI/ACCESS-ISSM/actions/runs/19481487026/job/55753757909?pr=28#step:16:1014).

Please let me know how to proceed to ensure the default variant is correctly respected *and* the dependencies are explicitly stated as you requested.

Thanks.

packages/issm/package.py
```diff
@@ -73,37 +69,43 @@ class Issm(AutotoolsPackage):
     # --------------------------------------------------------------------
     # Dependencies
     # --------------------------------------------------------------------
-    # Build-time tools
+
+    # Build-time & runtime dependencies
+    # --------------------------------------------------------------------
```
harshula: Please undo this change to the comment.

harshula (changes requested):
Thanks for your perseverance! Please fix those two comments and then squash-merge with a concise summary.

PR #350, opened 2025-11-06
Use numerical versions for CABLE

Whyborn:
Uses numerical versions for CABLE, and sets the `stable` branch to be preferred.

Also slight change in the license statement and add self as maintainer.

### DIFF
--- a/packages/cable/package.py
+++ b/packages/cable/package.py
@@ -16,13 +16,10 @@ class Cable(CMakePackage):
 
     maintainers("SeanBryan51", "Whyborn")
     
-    license("CSIRO Open Source Software License v1.0", checked_by="anton-seaice")
+    license("LicenseRef-CSIRO-Open-Source-Software-License-v1.0", checked_by="anton-seaice")
 
     version("stable", branch="main", preferred=True)
     version("2025.11.000", tag="2025.11.000", commit="15f639dd33dfb15819304332d72c2b405b51b85e")
-
-    def url_for_version(self, version):
-        return "https://github.com/CABLE-LSM/CABLE/tarball/{0}".format(version)
 
     variant(
         "mpi",


packages/cable/package.py
```diff
@@ -17,9 +14,10 @@ class Cable(CMakePackage):
     homepage = "https://github.com/CABLE-LSM/CABLE"
     git = "https://github.com/CABLE-LSM/CABLE.git"
 
-    maintainers("SeanBryan51")
+    maintainers("SeanBryan51", "Whyborn")
 
-    version("main", branch="main")
+    version("stable", branch="main", preferred=True)
+    version("2025.11.000", tag="2025.11.000", commit="15f639dd33dfb15819304332d72c2b405b51b85e")
```
anton-seaice: Just confirming, 2025.11.000 builds with both library = True and library = False

packages/cable/package.py
```diff
@@ -17,9 +14,10 @@ class Cable(CMakePackage):
...
 
     def url_for_version(self, version):
         return "https://github.com/CABLE-LSM/CABLE/tarball/{0}".format(version)
```
anton-seaice:
```suggestion
```

Shall we remove this? git versions are preferred

packages/cable/package.py
```diff
@@ -17,9 +14,10 @@ class Cable(CMakePackage):
     homepage = "https://github.com/CABLE-LSM/CABLE"
     git = "https://github.com/CABLE-LSM/CABLE.git"
 
-    maintainers("SeanBryan51")
+    maintainers("SeanBryan51", "Whyborn")
```
anton-seaice:
```suggestion

license("CSIRO Open Source Software Licence v1.0", checked_by="anton-seaice")

```

There's no SPDX code for the CSIRO license, so maybe listing nothing is better

packages/cable/package.py
Whyborn: > Just confirming, 2025.11.000 builds with both library = True and library = False

Is this a question or a statement? The library is a valued variant, and as long as you provide one of the valid values, it does build

anton-seaice:
`cable@stable%gcc@13.2.0~ipo+mpi build_system=cmake build_type=Release generator=make library=access-esm1.6 arch=linux-rocky8-x86_64` fails

but is ok with intel-classic

Whyborn:
Ah, it seems that the Intel precision specifiers `-r8 -i8` are not recognised by GCC.

anton-seaice (approved):
Thanks @Whyborn - you could add the library=access-esm1.6 builds to the CABLE CI as well - see https://github.com/CABLE-LSM/CABLE/tree/main/.github/build-ci/manifests

You might need another approver now!

Whyborn:
Yea, that's probably wise. Would've picked up this `fdefault-real-8` issue I had on the first attempt.

packages/cable/package.py
harshula: FYI, Spack upstream feedback on `url_for_version`: https://github.com/spack/spack/wiki/Telcon:-2025-09-02

CodeGat (dismissed):
The updates to the manifests all look good.

packages/cable/package.py
Whyborn: So this means we should define either `url` or `url_for_version`? Aren't these only useful if the version definitions do not provide a commit or tag?

packages/cable/package.py
Whyborn: @harshula My reading of this doesn't provide any useful information on licences which don't have a SPDX ID (of which the CSIRO license is one). The [CSIRO License](https://research.csiro.au/dap/licences/csiro-open-source-software-licence-based-on-mit-bsd-open-source-licence/) does say it's based on the MIT/BSD open source license, but is distinct due to some CSIRO clauses.

harshula: Hi @Whyborn , The first example in the documentation clearly states, `license("<SPDX Identifier HERE>")`. Would you like me to check with Spack upstream what we should do if there is no SPDX Identifier?

anton-seaice: I swear I searched for this before, but better luck this time!

In https://spdx.github.io/spdx-spec/v2.3/other-licensing-information-detected/

it says use:

```
 	"LicenseRef-"[idstring]
where
[idstring] is a unique string containing letters, numbers, . and/or -.
```

in this case 

`license("LicenseRef-CSIRO-Open-Source-Software-Licence-v1.0", checked_by="anton-seaice")`

might be better?

harshula (approved):
Hi @Whyborn , Please remember to squash all your commits and create a concise git summary.

Hi @anton-seaice , Do you want to create a separate PR for your CI changes? If not, you could Squash-Merge all the commits in this PR or use a Merge Commit after squashing all of @Whyborn 's commits.

anton-seaice:
I don't mind if you just use single commit with a Co-Author listed, or squash down to two commits (+ merge commit)

Issue #342, opened 2025-10-27
GCOM fails to build with GCC


Whyborn:
The GCOM package (or at least the most up to date version) fails with GCC. This causes CI failures for UM/UM7 builds (and probably those with UM dependencies), which shouldn't necessarily block merging as all of our release builds use Intel.

PR #370, opened 2026-01-29
Turn off gcc build-CI for access-esm1.6

anton-seaice:
Turns off access-esm1.6 %gcc builds which are failing, as noted in #342 

( I don't mess with access-esm1.5, due to https://github.com/ACCESS-NRI/access-spack-packages/issues/371)


.github/build-ci/manifests/access-esm1p6/intel.spack.yaml.j2
```diff
@@ -0,0 +1,13 @@
+# This manifest is used by default for packages without a specific manifest under .github/build-ci/manifests/PACKAGE/*.j2
+spack:
```
aidanheerdegen: Is it worth modifying this to say why a copy of the default is here? To suppress the default `gcc` build ...

aidanheerdegen (approved):
Looks good. 

CI isn't firing because there is no change to the spack packages in question.

If you wanted to check you could do a no-op commit on the two packages. 

If you squash the merge (which I'd recommend at this stage) then they'd have no affect.

PR #383, opened 2026-02-11
Add default CICE5 builds when using CMake

### LINKED ISSUE #382: Investigate invoking CMake multiple times in a spack package
Following on from https://github.com/ACCESS-NRI/build-cd/issues/333

For access-om2, we could like to deploy the model with 3 cice5 executables, all deployed within the one spack package. This is currently possible with the Makefile build, but in the interest of moving to making CMake the default, it needs investigating for CMake

One option is to modify the CMake build (in the cice5 repo) to support this and a second is to modify the spack package to run the CMake build multiple times. In this issue, the second will be investigated.

The CMakeBuild class is defined in spack-packages - see https://github.com/spack/spack-packages/blob/develop/repos/spack_repo/builtin/build_systems/cmake.py#L283-L285 for spack v1

There are three phases: `("cmake", "build", "install")` although it would be ok to bundle all the operations in these three phases into a single phases.

So to run the the build three times, it may be as simple as modifying the cmake_args function and then calling `super.cmake, super.build and super.install` in sequence

(I think this relies on the CMake targets being unique, so creating seperate build_dirs for each build might be needed for robustness)

anton-seaice:
This change adds support for doing multiple CMake builds within a spack package.

For model releases & deployments, we use the defaults in this package. For OM2 this produces three executables.

For development, optimisation work etc, the user can specify the grid & processor layout, but only one executable per spack spec.

All current behaviour is unmodified, except that setting `build_system=cmake` without providing the layout variants is now supported, and builds the default layouts.

In the unlikely case we changed the default build layouts such that the executable names were not unique, some builds would be missing.

```
ll .../spack/om2/release/linux-rocky8-x86_64_v4/oneapi-2025.2.0/cice5-2026.01.000-etrk3e4vjol5algldjnsorw2hy3piagf/bin

-rwxr-xr-x 1 as2285 tm70 18651432 Feb 11 13:11 cice_auscom_1440x1080_30x27.exe
-rwxr-xr-x 1 as2285 tm70 19013264 Feb 11 13:11 cice_auscom_3600x2700_40x30.exe
-rwxr-xr-x 1 as2285 tm70 17885064 Feb 11 13:11 cice_auscom_360x300_15x300.exe
```

closes #382

### DIFF
--- a/spack_repo/access/nri/packages/cice5/package.py
+++ b/spack_repo/access/nri/packages/cice5/package.py
@@ -7,13 +7,27 @@ from spack_repo.builtin.build_systems.cmake import CMakePackage
 from spack_repo.builtin.build_systems.makefile import MakefilePackage
 from spack.package import *
 
+# These are the default layouts, inc 3 executables for OM2
+# alternatively, supply the 5 layouts variant to produce 1 executable
+OM2_LAYOUTS = [
+        {"nxglob": "360", "nyglob": "300", "blckx": "15", "blcky": "300", "mxblcks": "1"},
+        {"nxglob": "1440", "nyglob": "1080", "blckx": "30", "blcky": "27", "mxblcks": "4"},
+        {"nxglob": "3600", "nyglob": "2700", "blckx": "40", "blcky": "30", "mxblcks": "12"},
+    ]
+ESM1P6_LAYOUTS = [
+    {"nxglob": "360", "nyglob": "300", "blckx": "30", "blcky": "300", "mxblcks": "1"},
+]
+
+
 def _int_validator(s):
     """Test a string variant is a valid integer """
-    if (s.isdigit() and int(s) > 0):
-        return True
-    else:
-        print(f"ERROR: {s} not a valid integer")
-        return False
+    if (s != "none"):
+        if (s.isdigit() and int(s) > 0):
+            return True
+        else:
+            print(f"ERROR: {s} not a valid integer")
+            return False
+
 
 class Cice5(CMakePackage, MakefilePackage):
     """The Los Alamos sea ice model (CICE) is the result of an effort to develop 
@@ -32,7 +46,7 @@ class Cice5(CMakePackage, MakefilePackage):
     version("access-om2", branch="master")
     version("access-esm1.6", branch="access-esm1.6")
 
-    # by default, keep existing processor layout defined in makefile
+    # by default, keep existing processor layout
     # if nxglob, nyglob, blckx, blcky, mxblcks are supplied, then spack will switch to cmake
     build_system("makefile", "cmake", default="makefile")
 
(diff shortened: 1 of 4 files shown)


packages/cice5/package.py
```diff
@@ -110,19 +120,58 @@ class Cice5(CMakePackage, MakefilePackage):
...
+
+        # do the build and install
+        for layout in layouts:
+            self.layout = layout
+            super().cmake(pkg, spec, prefix)
+            super().build(pkg, spec, prefix)
+            super().install(pkg, spec, prefix)
+
     def cmake_args(self):
+        # cmake is called during super().cmake()
```
anton-seaice:
```suggestion
        # cmake_args is called during super().cmake()
```

packages/cice5/package.py
```diff
@@ -110,19 +120,58 @@ class Cice5(CMakePackage, MakefilePackage):
...
+            for variant in OM2_LAYOUTS[0].keys()
+        ]):
+            if self.spec.variants["model"].value == "access-esm1.6":
+                layouts = ESM1P6_LAYOUTS
+            else:
+                layouts = OM2_LAYOUTS
+        else:
+            # to-do: test this
+            raise Error(f"All of {[OM2_LAYOUTS[0].keys()]} "
+                        "variants must be set together if any are set")
```
anton-seaice:
```suggestion
                        "variants must be set if any are set")
```

packages/cice5/package.py
```diff
@@ -110,19 +120,58 @@ class Cice5(CMakePackage, MakefilePackage):
...
+            for variant in layout_variants
+        ]):
+            layouts = list()
+            layouts.append(dict())
+            for variant in layout_variants:
+                layouts[0][variant] = self.spec.variants[variant].value
+        # else if no layout variants are available, use the defaults
+        elif all([
+            self.spec.variants[variant].value == 'none' 
+            for variant in OM2_LAYOUTS[0].keys()
```
anton-seaice:
```suggestion
            for variant in layout_variants
```

packages/cice5/package.py
```diff
@@ -110,19 +120,58 @@ class Cice5(CMakePackage, MakefilePackage):
...
+        ]):
+            if self.spec.variants["model"].value == "access-esm1.6":
+                layouts = ESM1P6_LAYOUTS
+            else:
+                layouts = OM2_LAYOUTS
+        else:
+            # to-do: test this
+            raise Error(f"All of {[OM2_LAYOUTS[0].keys()]} "
```
anton-seaice:
```suggestion
            raise Error(f"All of {layout_variants} "
```

spack_repo/access/nri/packages/cice5/package.py
```diff
@@ -110,19 +120,58 @@ class Cice5(CMakePackage, MakefilePackage):
...
+                        "variants must be set together if any are set")
+
+        # do the build and install
+        for layout in layouts:
+            self.layout = layout
+            super().cmake(pkg, spec, prefix)
+            super().build(pkg, spec, prefix)
+            super().install(pkg, spec, prefix)
+
     def cmake_args(self):
```
harshula: Consider moving `cmake_args()` before `cmake_build_install()` since it appears to be part of the initialisation step.

anton-seaice: cmake_args is called in super().cmake()

harshula: I noticed there's new function `set_layouts()`, I'd move `cmake_args()` before that too.

anton-seaice: If the goal is chronological order, set_layouts runs before cmake_args ?

harshula: No, that's not the reason for the suggestion. `cmake_args()` is a function commonly implemented in `CMakeBuilder` SPRs, so someone reading the SPR is likely to look for it and/or expect to see it.

anton-seaice: Thanks - done

spack_repo/access/nri/packages/cice5/package.py
```diff
@@ -7,13 +7,27 @@
 from spack_repo.builtin.build_systems.makefile import MakefilePackage
 from spack.package import *
 
+# These are the default layouts, inc 3 executables for OM2
+# alternatively, supply the 5 layouts variant to produce 1 executable
```
dougiesquire:
```suggestion
# alternatively, supply the 5 layout variants to produce 1 executable
```

dougiesquire (approved):
LGTM @anton-seaice. One nothing suggestion to prove I looked

anton-seaice:
Ready for final reviews

See https://github.com/ACCESS-NRI/ACCESS-OM2/pull/135 for builds on gadi and links to repro tests with the OM2

spack_repo/access/nri/packages/cice5/package.py
```diff
@@ -7,13 +7,27 @@
 from spack_repo.builtin.build_systems.makefile import MakefilePackage
 from spack.package import *
 
+# These are the default layouts, inc 3 executables for OM2
+# alternatively, supply the 5 layout variant to produce 1 executable
```
dougiesquire: 😉 

```suggestion
# alternatively, supply the 5 layout variants to produce 1 executable
```

dougiesquire (approved):

spack_repo/access/nri/packages/cice5/package.py
```diff
@@ -113,26 +127,91 @@ class Cice5(CMakePackage, MakefilePackage):
...
+    def set_layouts(self, pkg, spec, prefix):
+        """Layout of cice processors to use. If variants are set, use those. 
+        Otherwise, use defaults."""
+        layout_variants = OM2_LAYOUTS[0].keys()
+
+        # if all 5 layouts variants are available, set the layouts dict
+        if all([
+            self.spec.variants[variant].value != 'none' 
+            for variant in layout_variants
+        ]):
```
harshula, line 182: It would be clearer that this is a list comprehension if the square brackets are immediately at the start and end of the list comprehension.

spack_repo/access/nri/packages/cice5/package.py
```diff
@@ -113,26 +127,91 @@ class Cice5(CMakePackage, MakefilePackage):
...
+            layouts = [{variant: self.spec.variants[variant].value
+                for variant in layout_variants}]
+        # else if no layout variants are available, use the defaults
+        elif all([
+            self.spec.variants[variant].value == 'none' 
+            for variant in layout_variants
+        ]):
```
harshula, line 189: It would be clearer that this is a list comprehension if the square brackets are immediately at the start and end of the list comprehension.

harshula (approved):
Thanks! Great work.

PR #362, opened 2026-01-08
For vn13p1-am, use Github for config_root_path

penguian:
Closes #343 .
Note that vn13p8-am will need another issue, branch and pull request.

PR #385, opened 2026-02-17
Pull UKMO components from Github

Whyborn:
With the recent migration of the UKMO model components to Github, we can pull the auxiliary components (GCOM, Casim, Socrates and UKCA) from ACCESS-NRI forks of the codes, instead of the MOSRS mirror.

Tested against the previous release configuration (the one used in the [`release-n96e`](https://github.com/ACCESS-NRI/access-am3-configs/tree/release-n96e) configuration) in [this PR](https://github.com/ACCESS-NRI/ACCESS-AM3/pull/25). Shows bitwise identical results for the first few months in the NetCDF output.

Should we also change the [svn location](https://github.com/ACCESS-NRI/access-spack-packages/blob/49754fb3f3c8e84b73ce5cfcc592ca382a8ae5ae/spack_repo/access/nri/packages/um/package.py#L21) to a `git` location while we're here?

### DIFF
--- a/spack_repo/access/nri/packages/gcom/package.py
+++ b/spack_repo/access/nri/packages/gcom/package.py
@@ -13,21 +13,12 @@ class Gcom(Package):
     maintainers("scottwales", "paulleopardi")
 
     version("7.8", tag="vn7.8", commit="65c857cc3201833360ff62b285e49082378dae42")
-    version("7.9", tag="vn7.9", commit= "6319ae016dcadc192842a06178f3dbd21a8af64f")
-    version("8.0", tag="vn8.0", commit= "5a122e9e2147c9a7486a00b9588356eb65324af9")
-    version("8.1", tag="vn8.1", commit= "e061e2787bd643d9a65679f153fb8e3ffd9d3186")
-    version("8.2", tag="vn8.2", commit= "fd143bb38e21fe03c7150c0754852c80e15df3d4")
-    version("8.3", tag="vn8.3", commit= "b7b890a181d8e31e4e80b731b9f8ad9a6e1a8bed")
-    version("8.4", tag="vn8.4", commit= "f4fa92eb4af4f1e4cf9d608b441e3c96f77b6a6d")
-
-    # See 'fcm kp fcm:gcom.xm' for release versions
-    # version("7.8", revision=1147)
-    # version("7.9", revision=1166)
-    # version("8.0", revision=1181)
-    # version("8.1", revision=1215)
-    # version("8.2", revision=1251)
-    # version("8.3", revision=1288)
-    # version("8.4", revision=1386)
+    version("7.9", tag="vn7.9", commit="6319ae016dcadc192842a06178f3dbd21a8af64f")
+    version("8.0", tag="vn8.0", commit="5a122e9e2147c9a7486a00b9588356eb65324af9")
+    version("8.1", tag="vn8.1", commit="e061e2787bd643d9a65679f153fb8e3ffd9d3186")
+    version("8.2", tag="vn8.2", commit="fd143bb38e21fe03c7150c0754852c80e15df3d4")
+    version("8.3", tag="vn8.3", commit="b7b890a181d8e31e4e80b731b9f8ad9a6e1a8bed")
+    version("8.4", tag="vn8.4", commit="f4fa92eb4af4f1e4cf9d608b441e3c96f77b6a6d")
 
     variant("mpi", default=True, description="Build with MPI")
 
--- a/spack_repo/access/nri/packages/gcom/package.py
+++ b/spack_repo/access/nri/packages/gcom/package.py
@@ -8,18 +8,26 @@ class Gcom(Package):
     """
 
     homepage = "https://code.metoffice.gov.uk/trac/gcom"
-    svn = "file:///g/data/ki32/mosrs/gcom/main/trunk"
-
+    # svn = "file:///g/data/ki32/mosrs/gcom/main/trunk"
+    git = "https://github.com/ACCESS-NRI/gcom"
     maintainers("scottwales", "paulleopardi")
 
+    version("7.8", tag="vn7.8", "65c857cc3201833360ff62b285e49082378dae42")
+    version("7.9", tag="vn7.9", "6319ae016dcadc192842a06178f3dbd21a8af64f")
+    version("8.0", tag="vn8.0", "5a122e9e2147c9a7486a00b9588356eb65324af9")
+    version("8.1", tag="vn8.1", "e061e2787bd643d9a65679f153fb8e3ffd9d3186")
+    version("8.2", tag="vn8.2", "fd143bb38e21fe03c7150c0754852c80e15df3d4")
+    version("8.3", tag="vn8.3", "b7b890a181d8e31e4e80b731b9f8ad9a6e1a8bed")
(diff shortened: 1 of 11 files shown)


penguian:
Since we will now have the capability to do this for AM3, what is stopping us for doing it for the UM in general (e.g. rAM3, CM3)? Would the general capability need to be documented in a separate issue? 

See also https://github.com/ACCESS-NRI/access-spack-packages/pull/377 which is still using MOSRS.

spack_repo/access/nri/packages/um/package.py
```diff
@@ -64,13 +64,6 @@ class Um(Package):
     # String variants have their default values set to "none" here.
     # The real default is set by the model.
 
-    # Revision variants.
-    _rev_variants = (
-        "casim_rev",
-        "jules_rev",
-        "shumlib_rev",
-        "socrates_rev")
-
```
penguian, line 74: Does removing the "_rev" variants force all builds to use GitHub only? Will MOSRS-based builds still succeed?

penguian (commented):
Is the intent to change all UM builds to use GitHub only, or just the AM3 builds?

penguian:
If the intent is to change all UM builds, what testing has been done with rAM3?

spack_repo/access/nri/packages/um/package.py
Whyborn: I'm genuinely unsure. The logic in the spack package is too hard to follow. The existing AM3 `spack.yaml` did not specify revisions or variants for these components- did that mean it took the revision from the specified UM revision?

Whyborn:
> Is the intent to change all UM builds to use GitHub only, or just the AM3 builds?

I don't see why we wouldn't. I've confirmed bitwise equivalent results for AM3, so we should be able to get the same for other components. What are the advantages to maintaining support for MOSRS components?

> If the intent is to change all UM builds, what testing has been done with rAM3?

No testing has been done yet, but we don't need to update its `spack-packages` version yet.

Whyborn:
@penguian Do you know the history as to why GCOM got its own spack package, but the other UKMO components did not e.g. `socrates`, `ukca`? Should we be making spack packages for these components?

penguian:
@Whyborn GCOM is compiled separately as a library. The other "components" are cherry-picked for source code extraction by FCM and are compiled into a monolithic UM executable. Short of re-architecting the other components for standalone build, or refactoring the FCM make so that it works differently from the way it does now, the simplest thing to do was leave the FCM make as-is, and use its features while respecting its limitations.

penguian:
> > Is the intent to change all UM builds to use GitHub only, or just the AM3 builds?
> 
> I don't see why we wouldn't. I've confirmed bitwise equivalent results for AM3, so we should be able to get the same for other components. What are the advantages to maintaining support for MOSRS components?

At this point, there is no advantage in maintaining MOSRS support. The main "gotcha" that I see coming is that the UKMO builds from UM 14.X GitHub differently from the way that we build from UM 13.X, so when we get to UM 14.X we will probably need to revisit the Spack builds. Hopefully by that stage we can do away with FCM -- but only if FAB works better with GitHub.

It is probably wise to test building with rAM3 now to make sure that we don't have to make changes for rAM3 to work that will affect AM3 as well.

Whyborn:
> @Whyborn GCOM is compiled separately as a library. The other "components" are cherry-picked for source code extraction by FCM and are compiled into a monolithic UM executable. Short of re-architecting the other components for standalone build, or refactoring the FCM make so that it works differently from the way it does now, the simplest thing to do was leave the FCM make as-is, and use its features while respecting its limitations.

Ah ok, so it's the only one that's a true library.

> The main "gotcha" that I see coming is that the UKMO builds from UM 14.X GitHub differently from the way that we build from UM 13.X, so when we get to UM 14.X we will probably need to revisit the Spack builds. Hopefully by that stage we can do away with FCM -- but only if FAB works better with GitHub.

Are we ever planning to upgrade to UM14? Do you think it'd be possible and/or productive to just CMake-ify the UM stack?

penguian:
As for UM 14 or LFRic, ask @heidinett or Chermell Engel. They might have a better handle on what the Atmosphere WG intends to do with ACCESS-rAM3 in future, vs what ACCESS-NRI is doing to track the progress of Momentum Partnership regional workflows and RAL4. (For example, see https://code.metoffice.gov.uk/trac/rmed/query?group=status&milestone=RAL4 ).

spack_repo/access/nri/packages/um/package.py
penguian: I will do some testing today.

spack_repo/access/nri/packages/gcom/package.py
```diff
@@ -8,18 +8,17 @@ class Gcom(Package):
     """
 
     homepage = "https://code.metoffice.gov.uk/trac/gcom"
-    svn = "file:///g/data/ki32/mosrs/gcom/main/trunk"
-
+    # svn = "file:///g/data/ki32/mosrs/gcom/main/trunk"
```
penguian, line 11: If you want to remove a line, just remove it, otherwise explain in the comment why the line is commented out.

spack_repo/access/nri/packages/gcom/package.py
```diff
@@ -8,18 +8,17 @@ class Gcom(Package):
...
+    git = "https://github.com/ACCESS-NRI/gcom"
     maintainers("scottwales", "paulleopardi")
```
penguian, line 13: Should you add yourself to maintainers?

spack_repo/access/nri/packages/um/package.py
penguian: CI is failing with
```
Run . /opt/spack/share/spack/setup-env.sh
==> Error: cannot load package 'um' from the 'access.nri' repository: name '_rev_variants' is not defined
==> Created environment default in: /opt/environments/default
==> Activate with: spack env activate default
==> Activated default environment in /opt/environments/default
==> Error: cannot load package 'um' from the 'access.nri' repository: name '_rev_variants' is not defined
Error: Process completed with exit code 1.
```
It is probably best to revert your removal of the `_rev` variants so that the remainder of the CI check can proceed.

penguian: Also, it is may be better to migrate from MOSRS to GitHub gradually. I will need to think of how to do this.

penguian: For the `UM` Spack build, the main configuration elements are the `model/*/rose-app.conf` files. The Python code for the `UM` Spack package recipe starts from the appropriate `rose-app.conf` file corresponding to the version, modifies this config based on the variants provided by the Spack spec, exports the result to corresponding environment variables, and then runs `fcm make` using `fcm-make.cfg`. The `fcm make` run depends only on the environment variables and the contents of `fcm-make.cfg`. In particular, the `config_root_path` variable determines where 
```
include = $config_root_path/fcm-make/$platform_config_dir/um-$config_type-$optimisation_level.cfg$config_revision
```
points to. If (e.g.) `config_root_path` is left as `fcm:um.xm_tr`, then `fcm make` will use the UM trunk from the MOSRS mirror. See https://github.com/ACCESS-NRI/access-spack-packages/pull/362

penguian: At this point, I would recommend waiting until I have made the changes needed for #362 to be merged.

penguian (changes requested):
Wait for pull request #362.

penguian:
See #400

Whyborn:
Superseded by #400

PR #390, opened 2026-02-26
Touch to test pre-release

penguian:
Try running the pre-release build.

### DIFF
--- a/spack_repo/access/nri/packages/access_am3/package.py
+++ b/spack_repo/access/nri/packages/access_am3/package.py
@@ -10,7 +10,7 @@ from spack.package import *
 
 class AccessAm3(BundlePackage):
     """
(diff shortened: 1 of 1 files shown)


harshula:
Hi @penguian , This does not trigger a Pre-release build on Gadi. This triggers a CI build in a Docker container. You need to modify a Model Deployment Repository like https://github.com/ACCESS-NRI/access-am3 .

penguian:
> Hi @penguian , This does not trigger a Pre-release build on Gadi. This triggers a CI build in a Docker container. You need to modify a Model Deployment Repository like https://github.com/ACCESS-NRI/access-am3 

Thanks for the clarification. I must have missed that part of the doc, since I my responsibility is more on the building side than the releasing sise.

PR #395, opened 2026-03-05
issm: depend on c and fortran compiler

harshula:
* Prevent build failure due to:
  [spack cc]: Error: SPACK_CC_* variables not set: this usually means a missing `depends_on("c", type="build")` in package.py
  [spack cc]: Error: SPACK_FC_* variables not set: this usually means a missing `depends_on("fortran", type="build")` in package.py

### DIFF
--- a/spack_repo/access/nri/packages/issm/package.py
+++ b/spack_repo/access/nri/packages/issm/package.py
@@ -71,7 +71,9 @@ class Issm(AutotoolsPackage):
     # Dependencies
     # --------------------------------------------------------------------
     # Build-time tools
+    depends_on("c", type="build")
     depends_on("cxx", type="build")
+    depends_on("fortran", type="build")
     depends_on("autoconf", type="build")
     depends_on("automake", type="build")
     depends_on("libtool", type="build")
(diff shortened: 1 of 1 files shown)


harshula:
After this PR is merged, the ISSM team should fix:
```
==> issm: Executing phase: 'install'
==> Error: ValueError: ZIP does not support timestamps before 1980

/g/data/tm70/hj1810/test-v1.1.1-config/access-spack-packages/spack_repo/access/nri/packages/issm/package.py:262, in install:
        259                for src_path in py_files:
        260                    # Use only the filename inside the archive
        261                    arcname = src_path.split("/")[-1]
  >>    262                    zf.write(src_path, arcname=arcname)
```

CodeGat (approved):
Looks good. As for that timestamp error, I investigated this a while ago but it seems transient, to do with the `+py-tool` variant in the `package.py`. 

The version of python spack uses (`3.6.8`) is too old for `zipfile.ZipFile(..., strict_timestamps=False)`, so I'm not certain we have a great solution yet - hopefully the ISM team can help.

PR #396, opened 2026-03-06
`issm`: Clamp mtime of  sources zipped up by `+py-tools`

CodeGat:
See https://github.com/ACCESS-NRI/ACCESS-ISSM/actions/runs/22739327801

## Background

For some reason, `issm`s sources are sometimes given an `mtime` for unix epoch time. When attempting to zip them up with `python 3.6`s `zipfile` library, we get an error as `zipfile` can't zip things before 1980. Since we're on an old version of python before `zipfile.ZipFile(..., strict_timestamps=False)`, we have to do it ourselves. 

## The PR

* Manually clamp the mtime of the files to be > 1980, by modifting the `zipfile.ZipInfo` directly. We need to set some other aspects as well (namely, compression type and permissions).

## Testing

The build succeeded below without error!

### DIFF
--- a/spack_repo/access/nri/packages/issm/package.py
+++ b/spack_repo/access/nri/packages/issm/package.py
@@ -11,7 +11,7 @@ import zipfile
 import os
 import time
 
-ZIPFILE_MIN_DATE: tuple[int] = (1980, 1, 1, 0, 0, 0)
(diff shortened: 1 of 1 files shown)


justinh2002 (approved):

justinh2002:
LGTM!

PR #411, opened 2026-04-16
creating an issm release for production (non developmental)

justinh2002:
This PR enables compiler optimisation for production builds of ISSM.

Disables debugging (--with-debugging=0)
Uses optimised compiler flags (e.g. -O2/-O3) for improved performance
Removes development-oriented overhead

includes subsidiary update that mumps has been integrated into ISSM source and no longer needed as an external package: https://github.com/ISSMteam/ISSM/commit/6cb8efa360ddaf46ff57cee0942718883aef6774


aidanheerdegen:
> No description provided.

Can you please add some meaningful text description about the PR, the changes and intent behind them. A sentence or two is fine, but go to town if you want to

.github/build-ci/manifests/issm/gcc.spack.yaml.j2
aidanheerdegen, line 1: Now that you have intel and OneAPI tests you might want to consider renaming this to reflect that it is using `gcc`.

packages/issm/package.py
```diff
@@ -0,0 +1,331 @@
...
+
+    homepage = "https://issm.jpl.nasa.gov/"
+    git = "https://github.com/ACCESS-NRI/ISSM.git"
+
+    maintainers("justinh2002", "lawrenceabird")
+
+    # --------------------------------------------------------------------
+    # Versions
+    # --------------------------------------------------------------------
+    version("upstream", branch="main", git="https://github.com/ISSMteam/ISSM.git")
```
aidanheerdegen: You need to add `versions` for the tags you want to use for release, e.g. [2026.04.16](https://github.com/ACCESS-NRI/ISSM/releases/tag/2026.04.16)

justinh2002: good idea

aidanheerdegen: Not just a good idea, it's a requirement for our "working practice"

https://github.com/ACCESS-NRI/dev-docs/wiki/Spack#versioning-working-practice

harshula:
Hi @justinh2002 , Why does this PR contain a commit from @penguian ? Can you please rebase your branch on top of the latest `api-v2` branch?

justinh2002:
> Hi @justinh2002 , Which does this PR contain a commit from @penguian ? Can please rebase your branch on top of the latest `api-v2` branch?

done that now

harshula:
Hi @justinh2002 , You can now remove petsc and not rename mumps in this PR. Just copying mumps will work now as per https://github.com/ACCESS-NRI/access-spack-packages/pull/414 .

packages/issm/package.py
harshula, line 1: There shouldn't be an ISSM SPR in this location.

justinh2002:
[Scability_report___ISSM (2).pdf](https://github.com/user-attachments/files/28044897/Scability_report___ISSM.2.pdf)
Testing can be found here

spack_repo/access/nri/packages/issm/package.py
```diff
@@ -89,6 +97,11 @@ class Issm(AutotoolsPackage):
     # PETSc is used for linear algebra in all builds
     depends_on("petsc~examples+metis+mumps+scalapack")
```
harshula: Probably better to put the depends_on() inside a `with when("~production"):`

spack_repo/access/nri/packages/issm/package.py
```diff
@@ -141,6 +153,18 @@ def setup_build_environment(self, env):
             for var in ("CFLAGS", "CXXFLAGS", "FFLAGS", "LDFLAGS"):
                 env.append_flags(var, self.compiler.openmp_flag)
 
+        # Production build: optimized release flags
+        if "+production" in self.spec:
```
harshula: `if self.spec.satisfies("+production"):`

spack_repo/access/nri/packages/issm/package.py
```diff
@@ -141,6 +153,18 @@ def setup_build_environment(self, env):
...
+            for var in ("CFLAGS", "CXXFLAGS", "FFLAGS"):
+                env.append_flags(var, "-O3 -DNDEBUG -fPIC")
+            if self.spec.satisfies("%intel") or self.spec.satisfies("%oneapi"):
+                for var in ("CFLAGS", "CXXFLAGS", "FFLAGS"):
+                    env.append_flags(var, "-fp-model precise")
+            elif self.spec.satisfies("%gcc") or self.spec.satisfies("%clang"):
+                for var in ("CFLAGS", "CXXFLAGS", "FFLAGS"):
+                    env.append_flags(var, "-march=native -mtune=native")
```
harshula: Are you sure you want to set march/mtune to the platform used for compilation? @manodeep , What do we do in other models?

spack_repo/access/nri/packages/issm/package.py
```diff
@@ -173,7 +195,19 @@ def configure_args(self):
             f"--with-petsc-dir={self.spec['petsc'].prefix}",
         ]
         
-        # Automatic Differentiation specific options
+        # Production specific options
+        if "+production" in self.spec:
```
harshula: `if self.spec.satisfies("+production"):`

spack_repo/access/nri/packages/issm/package.py
```diff
@@ -173,7 +195,19 @@ def configure_args(self):
...
+            args += [
+                "--disable-debugging",
+                "--enable-development",
```
harshula: Did you want `--disable-development`

spack_repo/access/nri/packages/issm/package.py
manodeep: We do have arch flags in some SPRs (e.g, [UM7](https://github.com/ACCESS-NRI/access-spack-packages/blob/8cedd8f2f4a0fb1b5017c9d3d6a246fd74148589/spack_repo/access/nri/packages/um7/package.py#L148)) but I would not recommend `-march=native -mtune=native` since that fixes the runtime instructions to the capabilities on the compiling machine. Might not be too big of a deal on Gadi where login/compile nodes are cascadelake CPUs which aren't too far from the runtime sapphirerapids CPUs; however, I can imagine in the future if build-cd starts working on Nectar VMs, then those CPUs could be dramatically different worse performing than any CPUs we use in production.

spack_repo/access/nri/packages/issm/package.py
justinh2002: Yup, going to make it a `--disable-development` flag. all `--enable-development` does is it prevents MATLAB/Python scripts from being copied into $ISSM_DIR/bin; instead, dev workflows use the source-tree scripts directly. It doesn't have any performance deficit.

spack_repo/access/nri/packages/issm/package.py
justinh2002: I have used `if "+ad" in self.spec` and `if "+wrappers" in self.spec`.

spack_repo/access/nri/packages/issm/package.py
justinh2002: I will remove `-march=native -mtune=native`

spack_repo/access/nri/packages/issm/package.py
justinh2002: would you like me to change those too?

spack_repo/access/nri/packages/issm/package.py
justinh2002: I've been using a different syntax in this SPR

spack_repo/access/nri/packages/issm/package.py
harshula: Yes, please.

spack_repo/access/nri/packages/issm/package.py
harshula: My understanding is that you can use the `spack.yaml`'s `target` option to manipulate this flag.

spack_repo/access/nri/packages/issm/package.py
harshula: Please use `self.spec.satisfies()`.

spack_repo/access/nri/packages/issm/package.py
```diff
@@ -35,8 +35,11 @@ class Issm(AutotoolsPackage):
     version("main", branch="main")
     version("access-release", branch="access-release")
     version("access-development", branch="access-development")
-
-    version("2026.04.16", tag="2026.04.16", preferred=True)
+    
+    # same version string as the tag, so users can easily specify a specific tagged version if desired.
+    # This is the recommended way to specify versions for reproducibility.
+    version("2026.04.16", tag="2026.04.16")
+    version("2026.05.18", tag="2026.05.18", preferred=True)
```
harshula: The convention seems to be to put the oldest version at the bottom and the newest version on top.

spack_repo/access/nri/packages/issm/package.py
justinh2002: I will leave arch-tuning out of the SPR then

spack_repo/access/nri/packages/issm/package.py
justinh2002: as it appears with `--disable-development`, the if !DEVELOPMENT block is active and [make](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) tries to install Python files using ${ISSM_DIR}/src/m/archive/*.py. But ISSM_DIR is not set in the Spack build environment, so it expands to the broken path /src/m/archive/*.py.

`--enable-development` skips this entire block. Since the spack recipe already handles script installation via the +py-tools zip archive, --enable-development should be used in both build variants

spack_repo/access/nri/packages/m1qn3/package.py
harshula, line 1: Hi @angus-g , Is it OK if we delete the m1qn3 SPR from `access-spack-packages`?

.github/build-ci/manifests/issm/gcc.spack.yaml.j2
harshula: Hi @justinh2002 , Should this file be renamed to `gcc.spack.yaml.j2`?

justinh2002: @harshula , indeed it should

spack_repo/access/nri/packages/m1qn3/package.py
justinh2002: Hi @angus-g, for readability I have linked here the updates made to upstream ISSM to embed `m1qn3` into the source code and subsequent deletion of the package from all make files: https://github.com/ISSMteam/ISSM/commit/6cb8efa360ddaf46ff57cee0942718883aef6774

the SPR is no longer needed for ISSM compilation

angus-g: Yes of course, I don't mind.

harshula (approved):
Thanks! Please do a squash merge with a concise commit summary.

PR #416, opened 2026-05-05
Specify the precise version to CABLE ESM1.6 library

Whyborn:
The version of the CABLE library is intended to be frozen for ESM1.6. Future versions of CABLE are not expected to work with ESM1.6, as it features major revisions of the model internals. For this reason, specify the version of CABLE used in the ESM1.6 library CI.


Whyborn:
@ccarouge I added you here since there was still ongoing discussions as to whether the CABLE library is frozen, but I think we've settled on it being frozen, and any bug fixes being branched from that commit.

aidanheerdegen:
@Whyborn can you add superficial (formatting) changes to the UM7 and CABLE packages to trigger CI checks please?

Whyborn:
Hmm, looks like the tagged version for the library is off by 1 commit (although it's not meaningful for the library's purpose). The tagged commit is the commit prior to [this GCC fix](https://github.com/CABLE-LSM/CABLE/commit/d07ca7590a7c31866aa701354616055e45a0b7ee). This library is never used for anything but the Intel UM7 build- might be better to remove the GCC library CI?

aidanheerdegen:
All the tests pass, which is great, but you've triggered a build of the UM package, not the UM7 package.

> The tagged commit is the commit prior to [this GCC fix](https://github.com/CABLE-LSM/CABLE/commit/d07ca7590a7c31866aa701354616055e45a0b7ee). This library is never used for anything but the Intel UM7 build- might be better to remove the GCC library CI?

If you're confident it will *never* be used with GCC, then fine. Otherwise it can be good to keep it in to make sure we're not causing regressions against GCC, and it can always be good to test with another compiler to make sure there isn't anything compiler specific being done.

ccarouge:
I am not confident that the CABLE library for AM3 will never be compiled with gcc. But I'm guessing we can add the gcc file once we work on the AM3 library, is that the point? 

> looks like the tagged version for the library is off by 1 commit (although it's not meaningful for the library's purpose)

I'm not sure what that means. Are we using a different version of CABLE than we thought we were using?

Whyborn:
> Are we using a different version of CABLE than we thought we were using?

Sort of? The commit we've tagged is the one that builds correctly with Intel. There were two following commits which made the library build correctly with GCC.

> I am not confident that the CABLE library for AM3 will never be compiled with gcc. But I'm guessing we can add the gcc file once we work on the AM3 library, is that the point?

Since the libraries for different applications will be fundamentally different builds, they should be separate runs of build CI (or a CI task that builds all versions of the library). We can add the appropriate GCC manifest when the AM3 library is ready.

> All the tests pass, which is great, but you've triggered a build of the UM package, not the UM7 package.

Whoops, will fix that.

aidanheerdegen (approved):
LGTM

PR #420, opened 2026-05-15
CI: Use MCR Manifests Where Applicable

### LINKED ISSUE #419: CI: Use manifests from MCRs where defined
See related issue https://github.com/ACCESS-NRI/build-ci/issues/264

## Background

Currently, we have some duplication in manifests defined in MCRs, and manifests defined in `access-spack-packages`, which we need to unify. 

`build-ci` was updated to allow for a non-caller repos to provide manifests via `inputs.spack-manifest-repository[-ref]` in https://github.com/ACCESS-NRI/build-ci/pull/265.

This means that we should use manifests from MCRs where they exist.

## Implementation Details

The workflow will be updated with the following rules in mind:

1. SBRs are excluded from having CI manifests (in the exclude section of the matrix).
2. SPRs with MCRs: Duplication can be avoided by using the MCR CI manifests (in which we need to do a glob find in `.github/build-ci/manifests/*` to find individual manifests).
3. SPRs without MCRs: CI manifests will reside in access-spack-packages (if we can't find any manifests, we fallback to manifests defined in `access-spack-packages` `.github/build-ci/manifests/*/*`, and then finally the template manifests in `.github/build-ci/manifests/*`).
4. Consider how to disable "SPRs without MCRs" building with a particular compiler (outside of creating a specific directory in `.github/build-ci/manifests/PACKAGE`)?

Thanks @harshula for working through this with me!

CodeGat:
Closes #419
Required by https://github.com/ACCESS-NRI/spack-config/pull/95

## Background

Currently, we source manifests only from `access-spack-packages`. This means that there is potential for duplication or desyncronisation between MCRs and `access-spack-packages` manifests. 

Now, we will search for manifests for a given package via the following heuristic:

* Search for manifests under the MCRs `.github/build-ci/manifests` directory
* If none are found, search under `access-spack-packages` `.github/build-ci/manifests/PACKAGE` directory (for example, for packages that aren't managed directly by us, or don't have a git repo accessible to us)
* If none are found, search under `access-spack-packages` `.github/build-ci/manifests/*.spack.yaml.j2` for some very basic `gcc` and `intel` templates

## The PR

- **Move setup-matrix step to custom action**
- **Update generate-manifest-matrix action to search MCRs first**

## Testing

Done in this PR -  see https://github.com/ACCESS-NRI/access-spack-packages/actions/runs/27801357190?pr=420 for an example with MCRs (`cice4`, `cice5`, `access_mom6`) and one without (`datetime_fortran`)

## Other Notes

* This CI requires that MCRs have a `.github/build-ci/data/standard.json` file defined. Despite it being technically optional by `build-ci` itself, it would be quite a bit of extra code to conditionally add `inputs.spack-manifest-data-path` depending on the MCR, and the job name would be nearly unreadable (if they're not already!)


CodeGat:
Note that for repositories that do not use `run-self-hosted: false` (like CABLE in https://github.com/CABLE-LSM/CABLE/blob/main/.github/workflows/ci.yml#L24), we will need to check if that is set in the workflow and also add that to our matrix...

EDIT: Or do we? This is `ACCESS-NRI/access-spack-packages`, after all

CodeGat:
One last thing to do (outside of dropping the test commit...) is determining what to do with the existing `access-spack-packages` `.github/build-ci/manifests/PKG` for packages that have MCR manifests - maintainers may need to sync up manifests from `access-spack-packages` to their MCRs.

If there are manifests in `access-spack-packages` and in MCRs, the ones in `access-spack-packages` will be unused. And in that case...will there ever be a case where we would want to run both...?

aidanheerdegen:
> If there are manifests in `access-spack-packages` and in MCRs, the ones in `access-spack-packages` will be unused. And in that case...will there ever be a case where we would want to run both...?

I think running both would only be required if the purpose of the CI checks in `access-spack-packages` was different to the checks in the corresponding component repository. e.g. if `access-spack-packages` wanted to check every package against the most recent code versions of all the packages on which each package depends, or if `access-spack-packages` CI wanted to test against a matrix of compiler versions.

.github/build-ci/actions/generate-manifest-matrix/action.yaml
```diff
@@ -0,0 +1,124 @@
...
+        for pkg in $pkgs; do
+          # We want each manifest to have an associated injection of {{ package }} to $pkg
+          # Replace underscores with hyphens for the template value, as spack converts directory underscores to package name hyphens
+          template_value="${pkg//_/-}"
+          package_file="${{ inputs.packages-root-dir }}/$pkg/package.py"
+
+          ### Check for manifests from the packages associated GitHub repo, if it exists
+
+          # Extract OWNER/REPO directly when `git` points at GitHub.
+          repository=$(sed -nE "s|^\s*git\s*=\s*['\"]https://github\.com/([^/]+)/([^/'\"]+)['\"].*$|\1/\2|p" "$package_file" | head -n 1)
```
aidanheerdegen: When I run this in a shell the following packages have no associated GitHub URL
```
access_test_component
access_ww3
datetime_fortran
fcm
gcom
nci_intel_oneapi_mkl
um_createbc
um
```
Now some are fine, they are either external tools or use svn repositories, but `access_test_component` and `access_ww3` seem problematic.

.github/build-ci/actions/generate-manifest-matrix/action.yaml
```diff
@@ -0,0 +1,124 @@
...
+            # Some package repositories are private; use GH_TOKEN with `gh repo clone`.
+            if gh repo clone --no-upstream "$repository" "$cloned_repo_path" -- --depth 1 --filter=blob:none --quiet; then
+              if [[ -d "${cloned_repo_path}/.github/build-ci/manifests" ]]; then
+                # Use any manifests from the package's source repository first.
+                cd $cloned_repo_path
+                manifest_paths=$(find .github/build-ci/manifests -iname '*.j2' -type f -printf '%p ')
+                ref=$(git branch --show-current)
+                echo "Found manifests for $pkg under $repository at ref $ref: $manifest_paths"
+                cd - > /dev/null
+              fi
```
aidanheerdegen, line 125: Consider removing the repo at this point to ensure private repos are not leaked.

aidanheerdegen (commented):
I have questions

.github/build-ci/actions/generate-manifest-matrix/action.yaml
CodeGat: Hmm yes, I'll look into them.

`access_ww3` seems like it should fit the bill but doesn't - I'll update the regex
`access_test_component` is a bit troubling - currently we `sed` for the `git =` bit, because we can't load the `package.py` in python because we can't import the various spack imports we need without a `spack` installation. I suppose we could run it in a separate job that runs in a `build-ci-runner` container, just for this purpose...? Alternatively we could read the file via the `ast` library, but that's a lot of stuff.

CodeGat: Or just rename that one git variable 😆

.github/build-ci/actions/generate-manifest-matrix/action.yaml
CodeGat: Done in https://github.com/ACCESS-NRI/access-spack-packages/pull/420/commits/eaa572a5bb629465a64b739db2ede658fbd03b70

.github/build-ci/actions/generate-manifest-matrix/action.yaml
aidanheerdegen: I am big fan of work-arounds until they become technical debt. Might the same thing happen again? In which case it would fail silently. So if we did workaround this we should report when there are potential issues with matching the git repo address.

Can the python be loaded/imported but the spack imports mocked so it doesn't complain about the missing imports? If it isn't easy then I'd go with the work-around and add a comment to the code about the limitations and be noisy if there are patterns we should, but can't, match.

CodeGat: Whew! I've fixed this up. Now, the entire setup job runs in a `ghcr.io/access-nri/build-ci-runner` container (much like `build-ci` more generally), which allows us to use `spack python` to interrogate the packages `git` attributes (and interpolate them like in the case of `access_test_component`). This comes with the requirement that the use of the `generate-manifest-matrix` action must be in a container that contains `spack`. 

Done in https://github.com/ACCESS-NRI/access-spack-packages/pull/420/changes/eaa572a5bb629465a64b739db2ede658fbd03b70..05ca4ffc8ccb02cd8a496c45e0e19faf7e3efcb2

Tested in https://github.com/ACCESS-NRI/access-spack-packages/actions/runs/28004060089/job/82882170793?pr=420

.github/build-ci/actions/generate-manifest-matrix/action.yaml
```diff
@@ -0,0 +1,152 @@
...
+          # We want each manifest to have an associated injection of {{ package }} to $pkg
+          # Replace underscores with hyphens for the template value, as spack converts directory underscores to package name hyphens
+          template_value="${pkg//_/-}"
+          package_file="${{ inputs.packages-root-dir }}/$pkg/package.py"
+
+          ### Check for manifests from the packages associated GitHub repo, if it exists
+
+          # Extract git value from package.py (or an empty string if it doesn't exist),
+          # We use spack's python environment which has access to the package repository metadata
+          repository_url=$(spack python -c "import spack.repo; print(getattr(spack.repo.PATH.get_pkg_class('$pkg'), 'git', ''))")
```
aidanheerdegen, line 101: As discussed could use python [`urllib.parse` package](https://docs.python.org/3/library/urllib.parse.html) to extract the repository directly in this same code block.

CodeGat: Done in https://github.com/ACCESS-NRI/access-spack-packages/pull/420/commits/d84b0c08ac381d5d7098bfd2c8b071079e798a36

CodeGat:
As per a conversation with @harshula, we should still run non-template `access-spack-packages` manifests for a package even if there are MCR manifests. So the updated heuristic would be:

* Search for manifests under the MCRs `.github/build-ci/manifests` directory
* Search under `access-spack-packages` `.github/build-ci/manifests/PACKAGE` directory (for example, for packages that aren't managed directly by us, or don't have a git repo accessible to us)
* If none are found in either, search under `access-spack-packages` `.github/build-ci/manifests/*.spack.yaml.j2` for some very basic gcc and intel templates

CodeGat:
@harshula I have tested the above heuristic changed in https://github.com/ACCESS-NRI/access-spack-packages/pull/420/commits/ba43b0c2ff8c966c145967106be2f957522454c2 in https://github.com/ACCESS-NRI/access-spack-packages/actions/runs/28151601761, and it works!

aidanheerdegen (approved):
LGTM.

PR #421, opened 2026-05-18
Add UM_CBL preproc directive to build config

Whyborn:
The CABLE in JULES, which is built by the `um` spack package, requires the `UM_CBL` preprocessor directive with the recent update of the code (see JULES [29fac91](https://github.com/ACCESS-NRI/JULES/commit/29fac917d90d5133b6fe9b879dc3f9d9860f7757)). Should be a temporary addition, until CABLE is included as a library.

### DIFF
--- a/spack_repo/access/nri/packages/um/model/vn13p1-am/rose-app.conf
+++ b/spack_repo/access/nri/packages/um/model/vn13p1-am/rose-app.conf
@@ -2,8 +2,8 @@
 #meta=um-fcm-make/vn13.1
 
 [env]
-keys_atmos_extra=CABLE_UM_ISSUE24 CABLE_UM_ISSUE18 
-keys_scm_extra=CABLE_UM_ISSUE24 CABLE_UM_ISSUE18 
+keys_atmos_extra=CABLE_UM_ISSUE24 CABLE_UM_ISSUE18 UM_CBL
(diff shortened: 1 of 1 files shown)


penguian:
Is there a JULES issue or pull request related to JULES [29fac91](https://github.com/ACCESS-NRI/JULES/commit/29fac917d90d5133b6fe9b879dc3f9d9860f7757)? Is there an explanation of what the presence or absence of `UM_CBL` does?

penguian (commented):
See my comment to the pull request. I don't currently understand what `UM_CBL` does, and can't find a JULES issue or pull request related to it.

Whyborn:
@penguian The PR is [this closed PR](https://github.com/ACCESS-NRI/JULES/pull/324). It took all the CABLE standalone science code, and dumped it in the JULES `src/cable_science` directory, in preparation for bringing CABLE as a library.

Whyborn:
@penguian Is there any more information you need for this?

Whyborn:
@penguian Actually, I realised this may not be the best place to do this. It would be better to modify the source build config [here](https://github.com/ACCESS-NRI/UM/blob/AM3-dev/fcm-make/inc/um-atmos-common.cfg) so that both spack builds and in-suite FCM builds work.

PR #432, opened 2026-05-29
access_test & access_test_component: MPI is not optional

harshula:
* https://github.com/ACCESS-NRI/access-test-component/blob/19cbd388607dc3d1f4d989e1b3b1e3b73445d131/stub/CMakeLists.txt#L11
* We are using ACCESS-TEST to debug Spack concretizer issues, so we need to make sure the SPR is 100% accurate.

### DIFF
--- a/spack_repo/access/nri/packages/access_test/package.py
+++ b/spack_repo/access/nri/packages/access_test/package.py
@@ -19,9 +19,6 @@ class AccessTest(BundlePackage):
 
     version("latest")
 
-    variant("mpi", default=True, description="MPI build")
-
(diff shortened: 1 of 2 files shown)


aidanheerdegen (approved):
I can no longer recall the rationale for making MPI optional. If we never use or need that functionality then I agree this is the right approach.

On the other hand if we did need to have a non-MPI version then the CMake build should reflect this. Assuming we don't, I approve these changes.

harshula:
At this stage, we have to make sure that the SPRs accurately represent the current state of https://github.com/ACCESS-NRI/access-test-component/blob/main/stub/CMakeLists.txt .

PR #435, opened 2026-06-01
TEST: Multi-Node Self-Hosted Runners

CodeGat:
I've scaled the self-hosted cluster to 2 nodes, lets see how much it can handle...


CodeGat:
NOTE: It didn't work because required PVCs (buildcache, upstream) that were on one node were not on the other

PR #442, opened 2026-06-15
Add missing dependency on cmake with the version

harshula:
Spack upstream pointed out:
https://github.com/spack/spack-packages/blob/1ff98344f91f223211ad1f8b9c4858fb4c54e8fc/repos/spack_repo/builtin/build_systems/cmake.py#L169-L197

i.e. Spack will automatically add `depends_on("cmake", type="build")` when `"build_system=cmake"`, however, it does not contain a minimum CMake version. It may be safer to explicitly add it with the minimum CMake version.

### DIFF
--- a/spack_repo/access/nri/packages/access_fms/package.py
+++ b/spack_repo/access/nri/packages/access_fms/package.py
@@ -42,6 +42,7 @@ class AccessFms(CMakePackage):
     depends_on("cxx", type="build")
     depends_on("fortran", type="build")
 
+    depends_on("cmake@3.12:", type="build")
     depends_on("netcdf-c")
     depends_on("netcdf-fortran")
     depends_on("mpi")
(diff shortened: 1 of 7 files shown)


CodeGat (commented):
Assuming the checks go well, looks good. There are a bunch of different `cmake` versions for each package, where are they from? The associated MDRs?

harshula:
The CMake version comes from the source package's `CMakeLists.txt`.

CodeGat (approved):

anton-seaice (approved):
Looks good - spot checked some CMake versions and they are consistent

aidanheerdegen (approved):

aidanheerdegen:
Can we add a QA check to the CI that makes sure the CMake `depends_on` is set for all CMake packages?

harshula:
@CodeGat :point_up:

dougiesquire (approved):

CodeGat:
Yes, it's possible, but it seems like that would be it's own workflow that runs (probably outside the scope of this PR!)

PR #444, opened 2026-06-17
cice5: remove makefile support

harshula:
We have an opportunity to remove support for the legacy build system of CICE5 before the ACCESS-ESM1.6 Release. My understanding is that none of our released coupled models require CICE5 to be built via the legacy Makefile based build system.

### DIFF
--- a/spack_repo/access/nri/packages/cice5/package.py
+++ b/spack_repo/access/nri/packages/cice5/package.py
@@ -1,8 +1,9 @@
 # Copyright Spack Project Developers. See COPYRIGHT file for details.
 #
+# Copyright ACCESS-NRI
+#
 # SPDX-License-Identifier: (Apache-2.0 OR MIT)
 
-from spack_repo.builtin.build_systems import cmake
 from spack_repo.builtin.build_systems.cmake import CMakePackage
 from spack.package import *
 
@@ -36,7 +37,7 @@ class Cice5(CMakePackage):
     homepage = "https://www.access-nri.org.au"
     git = "https://github.com/ACCESS-NRI/cice5.git"
 
-    maintainers("harshula", "anton-seaice")
+    maintainers("anton-seaice", "harshula")
     license("BSD-3-Clause", checked_by="anton-seaice")
 
     version("stable", branch="master", preferred=True)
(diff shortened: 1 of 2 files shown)


spack_repo/access/nri/packages/cice5/package.py
```diff
@@ -46,16 +46,6 @@ class Cice5(CMakePackage, MakefilePackage):
     version("access-om2", branch="master")
     version("access-esm1.6", branch="access-esm1.6")
```
anton-seaice: Should we delete these lines too:
```
# TODO: the versions below can be removed once we are convinced they are not in use anywhere
version("access-om2", branch="master")
version("access-esm1.6", branch="access-esm1.6")

anton-seaice:
I might just psuh some refinements to the build-ci manifests ? There's now some duplication there that's non-essential

spack_repo/access/nri/packages/cice5/package.py
harshula: Your call. I'm happy to delete them if you think they are not needed any more.

anton-seaice: yes let's remove them

harshula:
Hi @anton-seaice , Leave the `build-ci` manifests as they are for now. We'll probably be deleting them soon to either move to using the source package's CI manifests or a centralised repository. Please make sure the CI manifests in the source package's repository is exactly what you need.

anton-seaice (dismissed):
Thanks Harshula

aidanheerdegen (dismissed):
Seems fine to me as long as @anton-seaice and @dougiesquire are happy (which it seems they are).

Can we have a brief summary of what this change means we no longer support, e.g. what models will no longer be supported "going-forward", and how might those models still be compiled (I assume `git clone` an older version of `access-spack-packages`).

harshula:
Hi @anton-seaice , Do any models require the CICE5 makefile build?

anton-seaice:
> e.g. what models will no longer be supported "going-forward", and how might those models still be compiled (I assume `git clone` an older version of `access-spack-packages`).

Right - so anything using ACCESS-OM2 <= 2026.02.002 or ACCESS-ESM1.6 <= 2025.11.000 (these were also all deployed using spack 0.22, so the ship has sailed in some respects)

This doesn't impact anything used in a `release-` branch for ESM1.6-configs, but that covers all currrent `release-` branches for OM2-configs. The CMake build for CICE5 is bit for bit with the old makefile build

aidanheerdegen (approved):
LGTM. Thanks.

PR #449, opened 2026-06-24
Install build-atmos/bin for access3 UM variant

### LINKED ISSUE #448: Install build-atmos/bin directory for access3 variant of UM
### Missing scripts for CM3/ESM3
The new ESM3 spack build currently does not install the `um-atmos`  and `um_script_functions scripts` required for running the model. The CM3 suite [executes these scripts](https://github.com/ACCESS-NRI/access-cm3-configs/blob/cafc9a9afb8dd6567c6299d28db16e675e92bfbb/app/um/rose-app.conf#L4) when setting off the model run. These scripts originate in the [UM source code](https://github.com/ACCESS-NRI/UM/blob/vn13.0_nuopc/bin/um-atmos), and are typically installed by FCM into the build-atmos/bin directory alongside the UM executable:

```
> ls build-atmos/bin
um-atmos  um-atmos.exe  um_script_functions
```

The new ESM3 spack build, which builds the `access3` variant of the UM, installs the `build-atmos/include` and `build-atmos/lib` directories, and skips the `build-atmos/bin` directory: https://github.com/ACCESS-NRI/access-spack-packages/blob/88271e7d4bcc178931287054af014b003523176a/spack_repo/access/nri/packages/um/package.py#L72-L100

As a result, the `um-atmos` and `um_script_functions` scripts aren't available to set off a CM3 run. 

### Suggested solution
Following discussion on Zulip, installing the `build-atmos/bin` directory for the `access3` variant should work as a solution. This will also install an extraneous UM executable, though this may still be useful in the future once AM3/CM3 etc code branches are merged.

blimlim:
Closes #448 

This PR modifies the UM spack package to also install the `build-atmos/bin` directory for the `access3` variant. This ensures that the `um-atmos` and `um_script_functions` scripts are installed, which are required for running ACCESS-CM3.

### DIFF
--- a/spack_repo/access/nri/packages/um/package.py
+++ b/spack_repo/access/nri/packages/um/package.py
@@ -68,10 +68,6 @@ class Um(UmBasePackage):
         When +access3, the package installs libum-atmos.a
         um-recon.exe is always installed.
         """
-
-        # List of executables to install
-        um_exe = ["recon", "atmos"]
-
         if self.spec.variants["access3"].value:
 
             # Create a pkgconf file for the um library
@@ -89,6 +85,9 @@ class Um(UmBasePackage):
         # Install executables and accompanying files into the prefix directory,
         # according to the directory structure of EXEC_DIR, as described in (e.g.)
         # https://code.metoffice.gov.uk/trac/roses-u/browser/b/y/3/9/5/trunk/meta/rose-meta.conf
+
+        # List of executables to install
+        um_exe = ["recon", "atmos"]
         for exe in um_exe:
             bin_dir = join_path(f"build-{exe}", "bin")
             build_bin_dir = join_path(self.build_dir(), bin_dir)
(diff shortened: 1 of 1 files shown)


anton-seaice (dismissed):
Looks good to me, you better check with @penguian

spack_repo/access/nri/packages/um/package.py
```diff
@@ -69,8 +69,8 @@ def install(self, spec, prefix):
         um-recon.exe is always installed.
         """
 
-        # List of executables to install, always install recon
-        um_exe = ["recon"]
+        # List of executables to install
+        um_exe = ["recon", "atmos"]
```
penguian: Move lines 72 and 73 to just before line 92. The intervening `if` statement does not refer to `um_exe`.

penguian (changes requested):

spack_repo/access/nri/packages/um/package.py
blimlim: Good idea, I've moved it in 7a9acda

PR #453, opened 2026-07-02
issm: set sticky=True for variants.

harshula:
Using `requires` instead of `conflicts` appears to fix these failures: https://github.com/ACCESS-NRI/access-spack-packages/actions/runs/28562927940

### DIFF
--- a/spack_repo/access/nri/packages/issm/package.py
+++ b/spack_repo/access/nri/packages/issm/package.py
@@ -35,7 +35,7 @@ class Issm(AutotoolsPackage):
     version("main", branch="main")
     version("access-release", branch="access-release")
     version("access-development", branch="access-development")
-    
+
     # same version string as the tag, so users can easily specify a specific tagged version if desired.
     # This is the recommended way to specify versions for reproducibility.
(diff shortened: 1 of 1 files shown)


spack_repo/access/nri/packages/issm/package.py
```diff
@@ -133,11 +139,8 @@ class Issm(AutotoolsPackage):
     # GCC 14 breaks on several C++17 constructs used in ISSM
     conflicts("%gcc@14:", msg="ISSM cannot be built with GCC versions above 13")
 
-    # +wrappers requires +py-tools to access the wrappers
```
justinh2002, line 136: Could you retain the annotations?

justinh2002 (changes requested):

spack_repo/access/nri/packages/issm/package.py
harshula: Hi @justinh2002 , Isn't the `msg` string a duplicate of the comment?

justinh2002: I see! missed that

justinh2002 (approved):

PR #466, opened 2026-07-29
Cable is dep only when +cable

Whyborn:
Applies Anton's suggestion, to only include CABLE as a dependency when the spec has `+cable`.

### DIFF
--- a/spack_repo/access/nri/build_systems/um_base.py
+++ b/spack_repo/access/nri/build_systems/um_base.py
@@ -181,7 +181,8 @@ class UmBasePackage(Package):
         when="+eccodes")
     depends_on("netcdf-fortran@4.5.2:", type=("build", "link", "run"),
         when="+netcdf")
-    depends_on("cable library='access3'", type=("build", "link", "run"))
+    depends_on("cable library='access3'", type=("build", "link", "run"),
+        when="+cable")
 
     phases = ["build", "install"]
 
--- a/spack_repo/access/nri/build_systems/um_base.py
+++ b/spack_repo/access/nri/build_systems/um_base.py
@@ -58,11 +58,15 @@ class UmBasePackage(Package):
     _bool_variants = (
         "DR_HOOK",
         "eccodes",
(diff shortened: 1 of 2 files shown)


anton-seaice (commented):
Looks correct to me - how did you test it ?

Whyborn:
Yet to test it- I can test an AM3 and rAM3 build, are we building CM3 with Spack yet?

anton-seaice:
There's only prototype CM3 spack builds , so don't worry about those

anton-seaice:
I think the main thing is to check that it concretises with and without CABLE correctly.

penguian (dismissed):
Looks good.

Whyborn:
I'm testing with an AM3 that should be with CABLE, and a RAM3 that should be without CABLE.

paul-leopardi (approved):
Ditto

Whyborn:
Investigating this uncovered a much deeper issue, that the `rose-app.conf` files are not being honoured at all. I'll do a temporary fix here, which is to set the default value for the `cable` variant to be `False` and work on the fundamental problem in another PR.

Whyborn:
Confirmed that this does successfully turn the `cable` variant off when we don't want it (see [this GCOM test PR](https://github.com/ACCESS-NRI/gcom/pull/5)), and does include it when we do explicitly ask for it (see [this AM3 PR](https://github.com/ACCESS-NRI/ACCESS-AM3/pull/41)).

I've requested a quick re-review @anton-seaice @paul-leopardi

anton-seaice (approved):
Looks ok to me:

https://github.com/ACCESS-NRI/ACCESS-AM3/actions/runs/30512282966/job/90774974742

has cable

https://github.com/ACCESS-NRI/gcom/actions/runs/30512011514/job/90773940737?pr=5

does not

Ill leave stylistic questions to Paul!

penguian (approved):
Looks good

paul-leopardi (approved):
Ditto

PR #469, opened 2026-07-31
Change default variant value in um_base to be more descriptive

Whyborn:
In the `um_base.py` build system, we link the rose build configuration options to string variants so that they can be set via the `spack.yaml`. The default value for these string variants is "none", which is misleading- it does not actually mean that specific configuration option will be set to either "none" or an empty string, it actually means that the default value from the model's respective `rose-app.conf` will be used. This can easily lead to confusion as to what the build is actually doing when inspecting the concretization (and very much did in my case).

If we change the default value for these variants to be "default", then the concretization for the build shows "<variant>=default", which is much better at conveying the information that the linked build configuration option is using the default values from the built in `rose-app.conf`.

I've confirmed that this has no effect on the AM3 and rAM3 builds, bar changing the appearance of the concretization.


Whyborn:
@paul-leopardi This is ready to go. An example of what the concretization of rAM3 looks like with this change:
```
^um@13.5~DR_HOOK~access3~cable+eccodes+mpi+netcdf COUPLER=default build_system=generic casim_ref=default casim_rev=default casim_sources=default
 compile_atmos=default compile_createbc=default compile_crmstyle_coarse_grid=default compile_pptoanc=default compile_recon=default compile_scm=default compile_sstpert_lib=default
 compile_wafccb_lib=default config_revision=default config_root_path=default config_type=default extract=default fcflags_overrides=default gwd_ussp_precision=default jules_ref=default
 jules_rev=default jules_sources=default land_surface_model=default ldflags_overrides_prefix=default ldflags_overrides_suffix=default ls_precipitation_precision=default
 mirror=default model=vn13p5-rns mpp_version=default openmp=default optimisation_level=default platagnostic=default platform_config_dir=default portio_version=default
 prebuild=default recon_mpi=default shumlib_ref=default shumlib_rev=default shumlib_sources=default socrates_ref=default socrates_rev=default socrates_sources=default
 stash_version=default thread_utils=default timer_version=default ukca_ref=default ukca_rev=default ukca_sources=default um_ref=default um_sources=default 
platform=linux os=rocky8 target=x86_64_v4 %c,fortran=intel@2021.10.0
```
No change in resulting executable.

paul-leopardi:
Sorry for my delay in reviewing your change. Could you please try using `"rose"` instead of `"default"`? It is shorter, more descriptive and less potentially confusing, especially in places like `spack.yaml`.

paul-leopardi (changes requested):
Try changing `"none"` to `"rose"` everywhere instead of  using`"default"`.

Whyborn:
Can we compromise and say "rose_default"? I don't think the term "rose" conveys the message that it falls back to default values.

Whyborn:
@paul-leopardi I've updated the variant defaults to be `rose_default`, as the temporary fix before the proper rejig of the SPR. Asssuming [this AM3 PR](https://github.com/ACCESS-NRI/ACCESS-AM3/pull/66) deploys, this is ready to go.

paul-leopardi (approved):
Approved

PR #476, opened 2026-08-12
Remove unnecessary preproc directives

Whyborn:
The `CABLE_UM_ISSUE24` preproc directive was removed when we merged with UM v13.8.


harshula:
Hi @Whyborn , What's the relationship of this PR to https://github.com/ACCESS-NRI/access-spack-packages/pull/475 ?

Whyborn:
I made the modifications to the wrong file in that one. It should only be removed from the `vn13p8-am` version of the configuration.

bschroeter (approved):

PR #479, opened 2026-08-18
Add um7 version 2026.04.002

blimlim:
This PR adds the latest um7 version, 2026.04.002, to the um7 spack package. @harshula fyi – this will be the first um7 version which supports gfortran.


dougiesquire (approved):
Thanks @blimlim. Commit and tag check out

Issue #387, opened 2026-02-24
Replace -xCORE-AVX2 with -mavx2


dougiesquire:
We currently have `-xCORE-AVX2` being set in a few of our SPRs (see below). These flags are CPU restrictive and have prevented users from building our models with Intel compilers on other HPC systems (e.g. [NeSI](https://forum.access-hive.org.au/t/installing-access-om2-on-nesi-new-zealand-supercomputer/5566) and [Archer2](https://forum.access-hive.org.au/t/building-access-om2-on-archer2-uk-hpc/5872)). The [recommendation](https://github.com/ACCESS-NRI/MOM5/issues/72) from the software transformation team (i.e. @manodeep) is to replace these flags with `-mavx2`, which will still allow AVX2 instructions but won't prevent running on other CPUs.

We are in the process of replacing `-xCORE-AVX2` with `-mavx2` where the flag is used in CMake build systems, but some changes to SPRs are also required:

- `mom5`: in the `MakefileBuilder` (cmake is default)
- `access-mocsy`: in the `MakefileBuilder` (cmake is default)
- `cice5`: in the `MakefileBuilder` (cmake exists, but makefile is still default)
- `cice4`: `MakefileBuilder` only
- `oasis3_mct`: `MakefileBuilder` only

Note, `oasis3_mct` and `cice5` actually set `-axCORE-AVX2` which does allow running on different CPUs, but @manodeep still recommends changing to `-mavx2`.

dougiesquire:
Note, we also have `-xCORE-AVX512` being set in a few SPRs. Mostly these are being set when building ACCESS-ESM1.5 (e.g. in `cice4`, `oasis3_mct`, `mom5`), but I think `um7` [uses `-xCORE-AVX512` for ACCESS-ESM1.6](https://github.com/ACCESS-NRI/access-spack-packages/blob/4d188611aa79e7c5d8cd9f8945bd93c1c52186bf/spack_repo/access/nri/packages/um7/package.py#L147).

Should we change this @manodeep?

dougiesquire:
I'm finding that replacing `-xCORE-AVX2` with `-mavx2` in the ACCESS-OM2 stack [degrades performance (~12%)](https://github.com/ACCESS-NRI/MOM5/issues/72#issuecomment-3962828661).

Should we just be removing these flags altogether and letting Spack set the architecture-specific stuff via target

[...]

PR #480, opened 2026-08-21
ACCESS-ESM1.6: add gcc support

harshula:
This PR adds preliminary support for building ACCESS-ESM1.6 with GCC.

### Testing

* https://github.com/ACCESS-NRI/ACCESS-ESM1.6/pull/207
* https://github.com/ACCESS-NRI/UM7/pull/223

### DIFF
--- a/spack_repo/access/nri/packages/um7/package.py
+++ b/spack_repo/access/nri/packages/um7/package.py
@@ -107,15 +107,6 @@ class Um7(Package):
         Set environment variables to their required values.
         """
         env.prepend_path("PATH", self.spec["fcm"].prefix.bin)
-        oasis3_incs = [
-                join_path(self.spec["oasis3-mct"].prefix.include, subdir)
-                for subdir in ["psmile.MPI1", "mct"]]
-        ideps = ["gcom4", "netcdf-fortran"]
-        if self.spec.satisfies(self._VERSION_WITH_CABLE):
-            ideps.append("cable")
-        incs = [self.spec[d].prefix.include for d in ideps] + oasis3_incs
-        for ipath in incs:
-            env.prepend_path("CPATH", ipath)
         # The gcom4 library does not contain shared objects and
         # therefore must be statically linked.
         env.prepend_path("LIBRARY_PATH", self.spec["gcom4"].prefix.lib)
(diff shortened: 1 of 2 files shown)


spack_repo/access/nri/packages/um7/package.py
```diff
@@ -156,13 +152,24 @@ def edit(self, spec, prefix):
...
+                join_path(spec["oasis3-mct"].prefix.include, subdir)
+                for subdir in ["psmile.MPI1", "mct"]
+        ]
+        incs_list = [spec[d].prefix.include for d in ideps] + oasis3_incs
+        incs = " ".join([f"-I{d}" for d in incs_list])
         libs = " ".join([self._get_linker_args(spec, d) for d in ldeps] + ["-lgcom"])
 
         opt_value = spec.variants["optim"].value
         EXE_NAME = self._exe_name(opt_value)
+        # TODO: What is the impact of LINUX_INTEL_COMPILER=linux_intel_compiler
```
aidanheerdegen: Could this TODO link to an issue? We've found in the `payu` codebase there were a lot of old TODOs that: weren't being fixed, no-one knew the context of.

If we have an associated issue it can contain context and potentially also get addressed.

aidanheerdegen: Linking just involves putting an issue number, e.g.  #387, in the comment.

MartinDix: There's no use of `LINUX_INTEL_COMPILER` anywhere in the code so it could be removed from both compiler configs.

spack_repo/access/nri/packages/um7/package.py
```diff
@@ -305,8 +314,68 @@ def edit(self, spec, prefix):
 tool::ld                                           mpif90
 tool::ldflags                                      {FOBLANK} -g -traceback {FDEBUG} -static-intel {libs}
         """
+        elif spec.satisfies("%gcc"):
+            FFLAGS = f"{incs}"
+            if opt_value == "debug":
+                FO = "-O0"
+                FTRACEBACK = ""
+                FDEBUG = ""
```
aidanheerdegen, line 322: Just noting there are some debug options available

https://gcc.gnu.org/onlinedocs/gcc-14.4.0/gfortran/Debugging-Options.html

I have no opinion on their value.

aidanheerdegen (approved):
Looks good to me. Just once comment.

spack_repo/access/nri/packages/um7/package.py
MartinDix: I think not worth worrying about for now.

MartinDix (approved):
OK with me.

PR #481, opened 2026-08-21
Add version 2025.11.001 to CABLE vers

Whyborn:
Adds the `2025.11.001` version for `cable`, which is the version which adds GCC support for the library build.


harshula (approved):
Thanks!

PR #486, opened 2026-08-27
Base `spack.include`s for `build-ci`

### LINKED ISSUE #484: Add top-level `spack.includes` default
## Background

See the parent issue, but essentially we are looking to have a tiered set of `spack` configuration for the purposes of deduplication, via git-based `spack.includes` sections in CRs.

This repository will house the top-level, most general defaults for CR manifests. Anything put here is overridable in CRs, too. 

## Implementation Details/Questions

* What things should we include? `spack.concretizer`? Compiler and target requirements?
* Should there be multiple sets of includes in this repository, as well? For example:

```
access-spack-packages/
└── .github/
    └── build-ci/
        └── includes/
            ├── base.spack.yaml (very general)
            ├── intel.spack.yaml (includes base.spack.yaml)
            └── oneapi.spack.yaml (includes base.spack.yaml)
```
* Where should we put it? Is `.github/build-ci/includes` a good directory?

CodeGat:
Closes #484

## Background

We are looking to remove common configuration from individual CR manifests via layered configuration via a `spack.include` section. 

The configuration in this PR is the top level of included configuration, with common configurations for compilers, concretizers, packages and repos. Any of the config in this repository is easily overridable at the CR level.

> [!NOTE]
> `build-ci@v4`-specific CI updates for ASP will be done in https://github.com/ACCESS-NRI/access-spack-packages/issues/485, this PR is specifically for CRs using `build-ci@v4` that consume these base includes, not for ASP to work with `build-ci@v4`

## Structure

Currently, we are looking at having a `base` level of configuration, which contains information on compilers, packages and repos. 

We will also have a `pr` level of configuration (which inherits `base`), that is special, specifically for the case where we have an ASP PR that calls CRs, which then include ASP configuration - we want it to use the ASP PRs version of `access-spack-packages`, not the default `api-v2`. 

As a common entrypoint for both CR and ASP PR, we have a `ci` level of config, which makes the decision on inheriting `base` or `pr` (which then inherits `base`).

## The PR

* Add `.github/build-ci/includes/base` for config common to all compilers
* Add `.github/build-ci/includes/pr` for config to be used when `access-spack-packages` is creating PRs
* Add `.github/build-ci/includes/ci` as a common entrypoint

## Testing

Testing includes in https://github.com/ACCESS-NRI/MOM6/pull/74


.github/build-ci/include/base/repos.yaml
```diff
@@ -0,0 +1,10 @@
+# These are sensible defaults for manifests as part of build-ci, and can be overridden
```
CodeGat: We clone these repos every time under the environment because it is quicker for `spack` to clone the repo fresh rather than fetch/unshallow/merge new changes as the image itself becomes more and more out of date

harshula: Hi @CodeGat , Can you please add the explanation as a comment in repos.yaml?

CodeGat: Done in https://github.com/ACCESS-NRI/access-spack-packages/pull/486/commits/18c0511c3ad88e6b49131a8768bc67b0498e3a7b

.github/build-ci/config/base/packages.yaml
```diff
@@ -0,0 +1,24 @@
...
+   - '@3.12.13'
+  openmpi:
+   require:
+   - '@5.0.8'
+
+  # Common base compiler versions
+  gcc:
+    # FIXME: Once intel-oneapi-compilers depends on %gcc@13.2.0, make this a hard requirement
+    prefer:
+    - '@13.2.0'
```
harshula, line 18: Should we try: require: one_of: ['@8.5.0', '@13.2.0', '@15.1.0']?

aidanheerdegen: Would the order matter?

aidanheerdegen (dismissed):

.github/build-ci/config/base/packages.yaml
harshula: Hi @aidanheerdegen , Yes, the order is taken into account during concretization.

CodeGat: Unfortunately we don't have `@15.1.0` in the image at the moment. But materially what is the difference between `require: one_of: [@8.5.0, @13.2.0]`  and `prefer: @13.2.0`, when we have only two gccs installed and `compiler_mixing: false`? Just on the surface they seem quite similar

CodeGat: And wouldn't we want to preference the newer compiler first in the `require: one_of` case?

harshula: Using `one_of` is descriptive and communicates what we are intending. My understanding is that we are having problems when `@13.2.0` is selected, so don't we want `@8.5.0` to be the first preference at the moment?

harshula: ~~Having said that, if there isn't reuse of these Spack packages and we get rid of the packages Spack environment in the future, it doesn't really matter.~~ This PR is not related to the upstream image build issues.

harshula (dismissed):

.github/build-ci/config/base/packages.yaml
CodeGat: Fair enough regarding the descriptiveness :grin:  In the majority of cases we want `@13.2.0` to be used over `@8.5.0` (for packages built, etc), it's only really the `oneapi` compiler in which we want to use `@8.5.0` at the moment, so I think in the first instance we should try `[@13.2.0, @8.5.0]` first

harshula: I just realised, I'm mixing up the upstream image build issues and this PR! Sorry about that!

CodeGat: No worries!! In any case, it didn't seem to be a happy chap with the `require.one_of` in this case, anyway: https://github.com/ACCESS-NRI/MOM6/actions/runs/34792334518/job/103818783940#step:20:245 as part of https://github.com/ACCESS-NRI/access-spack-packages/pull/486/commits/4fd3a2f72b4ef2979be96643fe022f2b7bfebe5a

harshula (approved):

.github/build-ci/config/base/packages.yaml
harshula: Hi @CodeGat , Thanks, drop that commit.

aidanheerdegen (approved):
I'm approving again, but it seems the same as last time I approved, which is fine, just checking I am not going insane.

.github/build-ci/config/base/packages.yaml
CodeGat: Dropped!

PR #499, opened 2026-09-09
Rewrite um build config

Whyborn:
Rewrites the UM build config to be more "Spack-ish". Attempts to reduce the amount of complexity in the system, remove "hidden" configuration options and make it more flexible with where to retrieve components from.

The major changes:
* The existing `model` directory which defines baseline configurations for each model are gone. All configuration is now done via the model's `spack.yaml`.
* Model components are pulled from Github via `<component>_ref` variants, and components are pulled from MOSRS via `<component>_rev` variants. This means it is possible to pull some components from MOSRS and some from Github. It is *not* possible to specify both a `_rev` and `_ref` variant for the same component. This means that when one of these variants say `none`, it truly means none- this variant (or the config option backing the variant) is not used at all in the build.
* The specific executables to compile are settable via the `config_type` variants. This was technically already possible, but not exercised- this removes the need for different packages for `um` and `um_createbc`. It can be a single package, and build all of `um-atmos`, `um-recon`, `um-createbc` and `um-scm` using the same source code which is generally desirable I think.
* Instead of relying on environment variables to set the location of the build config to use, it's written explicitly to the executable's build config. Each executable gets it's own build config that is inspectable.

The one current limitation is that the `createbc` executable doesn't support the same `optimisation_level` options as the other executables- it has only `debug`, `safe` and `high`, while all the others all include `rigorous`. This would cause a build failure if anyone specified `optimisation_level=rigorous` and included `createbc` in the multi-valued `config_type` variant.

The method of handling Github components as dynamic resources is unchanged.


spack_repo/access/nri/build_systems/um_base_new.py
```diff
@@ -0,0 +1,366 @@
...
+        "13.3": 118802,
+        "13.4": 120750,
+        "13.5": 123226,
+        "13.6": 124981,
+        "13.7": 127030,
+        "13.8": 128625,
+        "13.9": 130128,
+    }
+
+    maintainers("penguian")
```
aidanheerdegen, line 58: I would prefer to have more than one maintainer for redundancy. Given this work are you happy to add yourself @Whyborn?

harshula: Hi @Whyborn , If you do change that line, please remember to use `git rebase` to put your feature branch on top of the latest `api-v2` branch because of https://github.com/ACCESS-NRI/access-spack-packages/pull/498 .

